PR #838, opened 2014-09-04
Wizard: Fix creating folders in wrong dir

TingPing:
With this the log file is only created when needed, not earlier. Also no logs are made when supybot-wizard is ran.

I also removed automatic creating of all referenced dirs in conf.py. There is a *possibility* this could cause an issue but ```supybot``` always creates them on start and ```supybot-wizard``` creates them on finish. So users shouldn't be deleting random things while supybot runs.

EDIT: Ok missed a few places we still need to create folders..


ProgVal:
should I close the PR?

Issue #948, opened 2014-12-14
Decode hex codes displayed in facebook RSS feeds


iyx:
For some reason non-standard (foreign accents, etc.) characters are not displayed correctly on rss feeds of facebook pages. Instead hexadecimal codes are used, e.g.:

`&#xfa;` = ú
`&#x161;` = š

And I even noticed `&quot;` instead of `"` in a title.

Feed example: https://www.facebook.com/feeds/page.php?id=387324341361486&format=rss20

I have tested it with Python 2.6 and Python 2.7

As far as I know it only affects facebook feeds and it is most likely facebook problem, but would it be possible for rss plugin to convert this codes and display text correctly?

GLolol:
Looking at #949 it looks like this particular (buggy) RSS feed is no longer available. Closing accordingly.

PR #949, opened 2014-12-15
Parse headlines and titles through HTMLParser() incase of ascii XML/HTML encodings.

joulez:
This should address issue #948


joulez:
Thanks for the FB..FB.

PR #982, opened 2015-01-01
Time: refactor code, make tztime respect plugins.Time.format

GLolol:
In `tztime`:
- Import pytz on load, not every time the command is called.
- Respect `plugins.time.format`.
- Use `irc.error(e, Raise=True)` instead of `return` in error handling.
- Fix a typo in the command description.

In `time`:
- simplify the code used for getting the format configuration (the if statement doesn't appear necessary and the option can just be fetched using `msg.args[0]`).

### DIFF
--- a/plugins/Time/plugin.py
+++ b/plugins/Time/plugin.py
@@ -63,6 +63,11 @@ try:
 except ImportError:
     tzlocal = None
 
+try:
+    import pytz
+except ImportError:
+    pytz = None
+
 class Time(callbacks.Plugin):
     """This plugin allows you to use different time-related functions."""
     @internationalizeDocstring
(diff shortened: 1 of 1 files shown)


ProgVal:
by removing `optional('channel')`, you made it impossible to use the format of a channel outside this channel

GLolol:
The help never mentioned specifying a channel was an option, so I assumed that behavior wasn't needed. ,

**(time `[<format>]` `[<seconds since epoch>]`)** -- Returns the current time in <format> format, or, if <format> is not given, uses the configurable format for the **current channel**. If no <seconds since epoch> time is given, the current time is used.

GLolol:
Updated. `testDocstring (test_i18n.I18nTestCase)` is what's causing the latest test to fail.

PR #986, opened 2015-01-02
AutoMode: add l1on-es

JesuTinto:


ProgVal:
these are not the right file names
Run these commands, and then push again:

```
git mv locales/plugins_AutoMode_AutoMode-es.po plugins/AutoMode/locales/es.po
git mv locales/plugins_Google_Google-es.po plugins/Google/locales/es.po
```

Mikaela:
after running those commands which @ProgVal gave, run these two too:

```bash
git add -A # adds all files for commit
git commit --amend # edis the previous commit
git push --force # force pushes it even if it can be dangerous (it's not in this case)
```

ProgVal:
I indeed forgot to mention committing.
But it's not necessary to amend the commit, though

JesuTinto:
Well I'm going to read good Before posting the .po files 
Thank you Sincerely
Aaron

Issue #275, opened 2011-11-18
Badwords: Add a configuration option to disable op kixk


Limnoria:
Issue sent from #limnoria at freenode by ProgVal (registered as ProgVal)

PR #989, opened 2015-01-03
BadWords: bug fix, don't kick ops, remove unused import

GLolol:
- Remove unused `math` import.
- Don't kick halfop+ users or those with `channel,op`. Closes ProgVal/Limnoria#275.
- Fix a strange bug that causes the bot to kick on **everything** if the last defined bad word is removed (likely because `self.words` is empty and starts matching everything).

### DIFF
--- a/plugins/BadWords/plugin.py
+++ b/plugins/BadWords/plugin.py
@@ -29,7 +29,6 @@
 ###
 
 import re
-import math
 import time
 
 import supybot.conf as conf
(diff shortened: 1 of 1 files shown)


GLolol:
Once again `testDocstring` causing the build to fail, not me!

ProgVal:
Shouldn't you change the message in case both the bot and the user are halfops. The user is not immune, they just can't be kicked

PR #1002, opened 2015-01-13
[discussion] Karma: refactor message parsing/limit Karma to nicks in channel

GLolol:
Limiting Karma to nicks only prevents things such as "I'm learning C++" from triggering a Karma response, which is a bit ridiculous.

### DIFF
--- a/plugins/Karma/plugin.py
+++ b/plugins/Karma/plugin.py
@@ -246,32 +246,33 @@ class Karma(callbacks.Plugin):
     def _doKarma(self, irc, msg, channel, thing):
         inc = self.registryValue('incrementChars', channel)
         dec = self.registryValue('decrementChars', channel)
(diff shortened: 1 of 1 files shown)


GLolol:
...Uhh actually I realize that really breaks the tests. Is it possible to create dummy users in the testcases to make this work properly?

Mikaela:
> limit Karma to nicks in channel

I don't like this as many channels do `google++` etc meaning the service or something, not some nick in channel.

Also Karma defaults to not saying anything when Karma is increased/decreased.

GLolol:
Maybe it makes more sense as a config variable, but I don't see the value of being able to assign karma to long strings of text. Maybe confining it to nicks or one word would be better.

ProgVal:
There is `plugins.karma.allowUnaddressedKarma` for that.

PR #1005, opened 2015-01-16
Plugindownloader: case-desensitize repo names

GLolol:
This involves removing all capital letters from the repo list itself, since it's probably overkill to introduce a case-insensitive dict class for just one small list. Having to capitalize the repo name correctly is pretty tedious and annoying.


ProgVal:
I like the idea, but I prefer to keep the case of the keys.
So I implemented it in a different way in d85e09e3ed05ad9b64b5cd1b089fb6daa6c00772

GLolol:
That's awesome, I didn't know utils had that! :)

PR #1007, opened 2015-01-17
add queue arg to irc.reply() for using sendMsg instead of queueMsg

ddan39:
just like it says...

im not positive if it needs to be modified in other reply() functions, but it appears that the one i changed is the main one used. there is also a much simpler reply() in ReplyIrcProxy that uses queueMsg but i do not know when ReplyIrcProxy is used... if not ever from plugin then there is no point of adding arg to it

### DIFF
--- a/src/callbacks.py
+++ b/src/callbacks.py
@@ -836,12 +836,12 @@ class NestedCommandsIrcProxy(ReplyIrcProxy):
             else:
                 cb._callCommand(command, self, self.msg, args)
 
-    def reply(self, s, queue=None, noLengthCheck=False, prefixNick=None,
+    def reply(self, s, usequeueMsg=True, noLengthCheck=False, prefixNick=None,
               action=None, private=None, notice=None, to=None, msg=None):
         """
         Keyword arguments:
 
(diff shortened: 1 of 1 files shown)


GLolol:
Pull requests go to the testing branch, not master.

ddan39:
yea i know, i read that right after i did it. but i dont think i can undo a pull request myself.

joulez:
Queue needs to be explicted True to support current expectations. Also new keyword args should be appended to the end of the function declarations.

ddan39:
to the end!? it looks so nice at the beginning tho, and the line widths worked so well haha. will have to go to a 3rd line or shift it around...

*grumbles about how dumb the 79 max line width is*

ProgVal:
In case you want to open a new pull request:
* Sorry, but I don't like your variable names. You should rename `usequeueMsg` to `queueMsg` (it's more about what we actually do than what internal variable we use). You should also rename the `sendMsg` variable because the name is confusing, but I have no idea what name to use.
* As @joulez said, you should add new arguments at the end, because you would break plugins that use `prefixNick` as a positional argument

PR #1009, opened 2015-01-17
add usesendMsg=False to irc.reply() to use sendMsg() instead of queueMsg...

ddan39:
for plugin coding if you want to explicitly tell irc.reply() to use sendMsg instead of the default queueMsg

### DIFF
--- a/src/callbacks.py
+++ b/src/callbacks.py
@@ -836,8 +836,8 @@ class NestedCommandsIrcProxy(ReplyIrcProxy):
             else:
                 cb._callCommand(command, self, self.msg, args)
 
-    def reply(self, s, noLengthCheck=False, prefixNick=None,
-              action=None, private=None, notice=None, to=None, msg=None):
+    def reply(self, s, noLengthCheck=False, prefixNick=None, action=None,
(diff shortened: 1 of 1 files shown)


ProgVal:
https://github.com/ProgVal/Limnoria/pull/1007#issuecomment-70358272

ddan39:
changed most of that :)

i think usesendMsg is good arg now.. yea?

i agree about the sendMsg variable name, but dont know what better to use, and imo it doesnt matter much since it is local to that function

ProgVal:
> i think usesendMsg is good arg now.. yea?
Nope, same issue. You should use something like `queueMsg` or `sendImmediately`, which tells what the argument actually does, instead of referring to an internal name of the core

ddan39:
yes was thinking something just like that, sendNow or fast... but sendImmediately is even nicer, and actually what the doc string for sendMsg says it does :)

ddan39:
changed the arg name, but left sendMsg variable...  nothing really looked better. only idea was msgSender, but it just doesnt look nice :(

last idea is to do "sendMsg(msg) if sendImmediately else queueMsg(msg)"  at every spot that it does the sendMsg()...

PR #1012, opened 2015-01-18
Anonymous: remove dangerous example from README

Mikaela:
Anonymous tell uses normal privmsg which doesn't check that services are
U-lined and not random user pretending to be them.

The correct way to identify to NickServ after already being connected is
`ircquote nickserv identify username password`.

### DIFF
--- a/plugins/Anonymous/README.md
+++ b/plugins/Anonymous/README.md
@@ -4,38 +4,6 @@ messages in query too.
 
 ## Usage examples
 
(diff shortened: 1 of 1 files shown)


ProgVal:
How did we end up having that doc in the README of this plugin, actually?

Mikaela:
I guess it was added before Limnoria/Limnoria-doc started getting more updated.

PR #1017, opened 2015-01-19
src/irclib.py: request multi-prefix

Mikaela:
According to @nyuszika7h it's already supported, but not requested.

This is the last thing missing from IRCv3.1 base compatibility.


ProgVal:
Could you group these three messages?

```
 self.queueMsg(ircmsgs.IrcMsg(command='CAP', args=('REQ', 'account-notify extended-join multi-prefix')))
```

PR #1024, opened 2015-01-22
Update Debug to time of shipping as default plugin

Mikaela:
* Rewrite README.md based on `plugin help Debug`.
* Update string that I am not sure what it was, but it's not accurate anymore.
* add messages.pot & locales/.
    * `.gitignore` is there to avoid git removing the directory and it can be removed as translations are added. I don't feel like translating at the moment and this seems very difficult plugin to translate if this should even be translatable.

### DIFF
--- a/plugins/Debug/messages.pot
+++ b/plugins/Debug/messages.pot
@@ -1,0 +1,131 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR ORGANIZATION
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"POT-Creation-Date: 2015-01-22 18:54+EET\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"MIME-Version: 1.0\n"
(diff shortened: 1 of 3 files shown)


plugins/Debug/messages.pot
```diff
@@ -0,0 +1,131 @@
...
+"Content-Type: text/plain; charset=CHARSET\n"
+"Content-Transfer-Encoding: ENCODING\n"
+"Generated-By: pygettext.py 1.5\n"
+
+
+#: plugin.py:32
+#, docstring
+msgid ""
+"\n"
+"This is for debugging purpouses and you shouldn't load this plugin unless\n"
```
NormalRa: Purposes! :D

plugins/Debug/plugin.py
```diff
@@ -30,8 +30,8 @@
 ###
 
 """
-This is for jemfinch's debugging only.  If this somehow gets added and
-committed, remove it immediately.  It must not be released with Supybot.
+This is for debugging purpouses and you shouldn't load this plugin unless
```
NormalRa: Ditto.

plugins/Debug/plugin.py
```diff
@@ -30,8 +30,8 @@
 ###
 
 """
-This is for jemfinch's debugging only.  If this somehow gets added and
-committed, remove it immediately.  It must not be released with Supybot.
+This is for debugging purposes and you shouldn't load this plugin unless
+Supybot developer requests you to do so for debugging some issue.
```
GLolol: There's some minor grammar things with this sentence, try: "This plugin is for debugging purposes only and you shouldn't load it unless a Supybot developer requests you to debug some issue."

Issue #1028, opened 2015-01-24
Failing to connect to freenode


Mikaela:
The latest lines from messages.log:

```
TypeError: expected bytes, not str
ERROR 2015-01-24T18:14:29 supybot Exception id: 0x7e15a
WARNING 2015-01-24T18:15:02 supybot Error message from freenode: Closing Link: vep.mikaela.info (Connection timed out)
INFO 2015-01-24T18:15:02 supybot Reconnecting to freenode.
INFO 2015-01-24T18:15:02 supybot Connecting to chat.freenode.net:7000.
INFO 2015-01-24T18:15:03 supybot freenode: Server acknowledged capability 'account-notify'
INFO 2015-01-24T18:15:04 supybot freenode: Server acknowledged capability 'extended-join'
INFO 2015-01-24T18:15:05 supybot freenode: Server acknowledged capability 'multi-prefix'
INFO 2015-01-24T18:15:06 supybot freenode: Server acknowledged capability 'sasl'
INFO 2015-01-24T18:15:08 supybot freenode: Authenticating using SASL.
ERROR 2015-01-24T18:15:10 supybot Uncaught exception in Irc.feedMsg:
Traceback (most recent call last):
  File "/home/users/mikaela/.local/lib/python3.2/site-packages/supybot/log.py", line 361, in m
    return f(self, *args, **kwargs)
  File "/home/users/mikaela/.local/lib/python3.2/site-packages/supybot/irclib.py", line 854, in feedMsg
    method(msg)
  File "/home/users/mikaela/.local/lib/python3.2/site-packages/supybot/irclib.py", line 1009, in doAuthenticate
    private_key.sign(base64.b64decode(msg.args[0]))).decode('utf-8')
  File "/usr/lib/python3.2/base64.py", line 83, in b64decode
    raise TypeError("expected bytes, not %s" % s.__class__.__name__)
TypeError: expected bytes, not str
ERROR 2015-01-24T18:15:10 supybot Exception id: 0x7e15a
WARNING 2015-01-24T18:15:42 supybot Error message from freenode: Closing Link: vep.mikaela.info (Connection timed out)
INFO 2015-01-24T18:15:42 supybot Reconnecting to freenode.
INFO 2015-01-24T18:15:42 supybot Connecting to chat.freenode.net:7000.
INFO 2015-01-24T18:15:44 supybot freenode: Server acknowledged capability 'account-notify'
[...]

PR #1029, opened 2015-01-24
irclib.py: fix an encoding error with Python 3 and SASL ECDSA-NIST256P-CHALLENGE

GLolol:
Temporary fix for #1028, though this may cause ECDSA-NIST256P-CHALLENGE to fail entirely (blocking the bot from connecting is a more severe issue).


ProgVal:
Doesn't it break Python 2 support?
(If yes, add somes lines like: `if sys.version_info[0] < 3: m = msg.args[0] else: m = msg.args[0].encode()`)

GLolol:
Python 2 doesn't seem to care about `encode()` when everything is ASCII, but I don't know the nature of ECDSA and what it sends.

PR #1035, opened 2015-01-31
ShrinkUrl: use httpUrl instead of url in wrap()

GLolol:
This allows things like '`x0 overdrive.pw`' to work properly, instead of requiring the command to be typed as '`x0 http://overdrive.pw'`. Most people won't be using URL shorteners for things that aren't `http(s)://` links anyways, so I think it should be safe to assume `http://` here.


ProgVal:
are you sure it actually makes the scheme optional?

GLolol:
```
<+GLolol> `x0 overdrive.pw
<@Lily> http://x0.no/iz5o
```

PR #1036, opened 2015-01-31
Network: default 'whois' to current nick if not specified

GLolol:
This is mostly a change for convenience, allowing easy whoising of yourself in a channel. 

However, one side effect of this is that: whoising a person who's nick = the current network name will require specifying the person's name twice. 

Running `whois <network>` will try to whois your current nick on network `<network>`.

### DIFF
--- a/plugins/Network/plugin.py
+++ b/plugins/Network/plugin.py
@@ -310,10 +310,12 @@ class Network(callbacks.Plugin):
         # The double nick here is necessary because single-nick WHOIS only works
         # if the nick is on the same server (*not* the same network) as the user
         # giving the command.  Yeah, it made me say wtf too.
+        if nick is None:
+            nick = msg.nick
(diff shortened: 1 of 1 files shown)


ProgVal:
I don't understand this: “the current network name will require specifying the person's name twice. ”

GLolol:
Well, if someone has the nick "freenode" on a network called "freenode", one would have to call `whois freenode freenode` (`whois freenode` whoises yourself on that network).

ProgVal:
so that's not a regression?

ProgVal:
ping

GLolol:
It might be but it's somewhat rare for someone to have the same nick as a network name.

GLolol:
Actually, I changed my mind. It's probably not worthwhile to change the entire command's behavior for something as small as this.

PR #1037, opened 2015-01-31
Admin.channels: call irc.reply() with private=True instead of requiring privacy

GLolol:
When `irc.reply()` is called with `private=True`, the bot will force a message in PM (or notice if configured to do so). IMO this makes more sense than forcing `channels` to be called in PM, as they both the same effect of preventing information leaks.


Mikaela:
:+1: and the test fail seems to be unreleated

PR #1041, opened 2015-02-05
Admin/Owner: substitute $version in quit and part messages (also make part messages configurable)

### LINKED ISSUE #391: There isn't a way to set default part message with a configuration variable.
The only configuration variable to set about default messages is for quitting, supybot.plugins.Owner.quitMsg

GLolol:
Closes #391.

This also makes the default part message the Supybot version instead of the nick's caller. Quit messages are optional anyways for IRC, so just having someone's nick inside a part message isn't very intuitive or meaningful. When `config plugins.admin.partmsg` is blank, simply no part message will be used.

edit: This also merges the changes in #1043 and supports other substitutions too (`$nick`, etc.).

### DIFF
--- a/plugins/Admin/config.py
+++ b/plugins/Admin/config.py
@@ -43,8 +43,11 @@ def configure(advanced):
 
 
 Admin = conf.registerPlugin('Admin')
-# This is where your configuration variables (if any) should go.  For example:
-# conf.registerGlobalValue(Admin, 'someConfigVariableName',
-#     registry.Boolean(False, """Help for someConfigVariableName."""))
+
+conf.registerChannelValue(Admin, 'partMsg',
+    registry.String('%version%', _("""Determines what part message should be
+        used by default. If the part command is called without a part message,
+        this will be used. If this value is empty, then no part message will
+        be used (they are optional in the IRC protocol).""")))
 
 # vim:set shiftwidth=4 softtabstop=4 expandtab textwidth=79:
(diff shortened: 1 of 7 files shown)


ProgVal:
Could you use template strings instead? https://docs.python.org/3/library/string.html#template-strings

ProgVal:
Actually, this may be even better: https://github.com/ProgVal/Limnoria/blob/master/src/ircutils.py#L668

GLolol:
Would it be appropriate to add another item in the vars dict for this? (`version: 'Supybot %s' % conf.version`)

I'll take a look at it later today.

ProgVal:
absolutely

ProgVal:
`ircmsgs.ping`?

GLolol:
The `ircutils.standardSubstitute` function expects a msg object (e.g. it expands `$nick` to `msg.nick`), but Ctrl-C at console isn't an IRC event and thus doesn't have one.

ProgVal:
I think it would be better to have an altered version of standardSubstitute that can be used without a msg object

GLolol:
Where should such a function be placed?

ProgVal:
Actually, I made a change to `standardSubstitute` so it now accepts `None` as `irc` and `msg`: https://github.com/ProgVal/Limnoria/commit/cb6669015e4dfd113f3daf6e3797708d7ce307c1

Could you update your patch to handle that?

GLolol:
Done.

Mikaela:
What is the current status of this pull request?

PR #1043, opened 2015-02-05
Owner: substitute %version% for the Supybot version in plugins.Owner.quitmsg

### LINKED ISSUE #847: Variables in quit/part messages
It would be nice if I could set generic quit/part messages like `Supybot 0.83.4.1+limnoria 2014.08.30` which would always be current version. Output of `misc version` would be a little too much and I am not sure if Python version from `supybot --version` is too much.

ZNC has `%version%` and WeeChat `%v` and both get expanded to VERSION.

*Part message with config (#391) isn't implemented yet.*

GLolol:
Closes #847. This also changes the default quit message to the Supybot version. 

Rephrasing my argument on #1041, I feel that having a person's nick in a default quit message isn't meaningful at all. If `plugins.owner.quitmsg` is empty, only then will the caller's nick be used.

### DIFF
--- a/plugins/Owner/config.py
+++ b/plugins/Owner/config.py
@@ -46,10 +46,10 @@ conf.registerGlobalValue(Owner, 'public',
     registry.Boolean(True, """Determines whether this plugin is publicly
     visible."""))
 conf.registerGlobalValue(Owner, 'quitMsg',
-    registry.String('', """Determines what quit message will be used by default.
+    registry.String('%version%', """Determines what quit message will be used by default.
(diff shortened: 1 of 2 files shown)


ProgVal:
Same comments as for https://github.com/ProgVal/Limnoria/pull/1041

GLolol:
#1041 was updated to merge these changes as well, closing accordingly. (`src/ircutils.py` was updated to fit both of these PRs, so having two of them can break if one is merged and the other one isn't).

PR #1047, opened 2015-02-06
src/i18n.py: mention supported languages

### LINKED ISSUE #1046: supybot.language: how about saying what values are suggested?
instead of "valid values are things like" there could be "Currently supported are:".

Mikaela:
Closes #1046

I used Admin for sources of what languages are supported as it's usually
recommended to start translating from there (first plugins, then core).


Mikaela:
feel free to quote then. en should probably be there, but I didn't think about it and I am not sure if I am going to continue doing anything to limnoria

PR #1051, opened 2015-02-10
Seen: prevent KeyError when calling commands on channels that the bot is not in

GLolol:
This fixes a bug introduced by 38ff1a1137.


ProgVal:
Looks like there is a lot of code duplication in this plugin.
I'll handle it later.

PR #1053, opened 2015-02-11
Aka: prevent DoS attacks with recursive Akas

### LINKED ISSUE #526: Aka thing broken again...
aka add foo foo $1$1$1$1$1$1 $1$1$1$1$1$1$\* $1$1$1$1$1$1$\* causes a fork bomb....

GLolol:
This change makes Aka keep track of all aliases (tokens) called during the expansion of an Aka, which aims to block most recursion-based DoS attempts. (Really fixes #526)

It will catch and terminate execution when of an Aka when either:

1. The same command & arguments are called multiple times during expansion of one Aka.
    - For example: `aka add test test` will call the `test` command endlessly if `test` is called once.

2. The amount of Akas called in expansion exceeds the limit set in `supybot.plugins.Aka.recursionLimit` (new variable). This can be used to block attacks such as:
    - `aka add test test $1$1$1$1$*`, where the command is recursive but the arguments change & grow infinitely large over time.
    - `aka add test test2`, `aka add test2 test`: any situation where a group of aliases are chained to call each other without stopping.

### DIFF
--- a/plugins/Aka/config.py
+++ b/plugins/Aka/config.py
@@ -55,6 +55,8 @@ conf.registerGlobalValue(Aka, 'maximumWordsInName',
     registry.Integer(5, _("""The maximum number of words allowed in a
     command name. Setting this to an high value may slow down your bot
     on long commands.""")))
-
-
+conf.registerGlobalValue(Aka, 'recursionLimit',
+    registry.NonNegativeInteger(8, _("""Determines the maximum level of Aka recursion
+    (Akas calling other Akas) allowed. This can help effectively stop DDoS attempts that
+    call one or more Akas in a loop. If set to zero, this check is disabled.""")))
 # vim:set shiftwidth=4 tabstop=4 expandtab textwidth=79:
--- a/plugins/Aka/plugin.py
+++ b/plugins/Aka/plugin.py
@@ -377,6 +377,7 @@ class Aka(callbacks.Plugin):
         self.__parent = super(Aka, self)
         self.__parent.__init__(irc)
         self._db = AkaDB()
+        self.akasUsed = []
 
     def isCommandMethod(self, name):
         args = name.split(' ')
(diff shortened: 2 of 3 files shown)


GLolol:
FYI the [last test fail](https://travis-ci.org/ProgVal/Limnoria/jobs/50291767) was due to a network error at Travis (pip failed to download something).

Mikaela:
Did Travis add the option to allow commit authors to restart builds yet?

Also, does this also affect Alias?

GLolol:
> Did Travis add the option to allow commit authors to restart builds yet?

I don't see a button for that.

> Also, does this also affect Alias?

It doesn't. I haven't used Alias for months personally and the two are separate plugins anyway (it would be a separate PR).

Mikaela:
>> Did Travis add the option to allow commit authors to restart builds yet?
> I don't see a button for that.

yes, it appears that https://github.com/travis-ci/travis-ci/issues/887 is still open and closed.

ProgVal:
1) Don't use a list as you only want to make searches in `akasUsed`.

2) I don't think this is the right idea. `akasUsed` is global, so the behavior won't be deterministic and `akasUsed` may be reset by another running Aka.
I suggest that you use `IrcMsg.nested` instead.

GLolol:
> 1) Don't use a list as you only want to make searches in akasUsed.

It also uses the length of the list.

> 2) I don't think this is the right idea. akasUsed is global, so the behavior won't be deterministic and akasUsed may be reset by another running Aka.

I did some debugging on my own and couldn't reproduce this. It should only clear when the first Aka is done fetching.

> I suggest that you use IrcMsg.nested instead.

Did you mean `irc.nested`? I don't know if that'll work because Akas can call each other and not be directly using nested commands.

ProgVal:
> It also uses the length of the list.
Works with sets too.

> I did some debugging on my own and couldn't reproduce this. It should only clear when the first Aka is done processing.
I know that it's hard to reproduce. But someone may be able to use it anyway.

> Did you mean `irc.nested`? I don't know if that'll work because Akas can call each other and not be directly using nested commands.
Yes. So we should increment the nesting when calling akas too.

GLolol:
Oh, that makes sense.

ProgVal:
Should I close this PR?

PR #1056, opened 2015-02-21
Network.whois: simplify message when user isn't on any public channels

GLolol:
The current message ("xyz isn't on any non-secret channels or is using a channel-list hiding umode.") is long and awkwardly worded. 

This also removes an extra period from showing up at the end of the output.

### DIFF
--- a/plugins/Network/plugin.py
+++ b/plugins/Network/plugin.py
@@ -243,7 +243,7 @@ class Network(callbacks.Plugin):
                     L.append(format(_('is on %L'), normal))
         else:
             if command == 'whois':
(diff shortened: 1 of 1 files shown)


plugins/Network/plugin.py
```diff
@@ -243,8 +243,7 @@ def do318(self, irc, msg):
                     L.append(format(_('is on %L'), normal))
         else:
             if command == 'whois':
-                L = [_('isn\'t on any non-secret channels or is using a '
-                    'channel-list hiding umode.')]
+                L = [_('isn\'t on any publically visible channels')]
```
Mikaela: Shouldn't this be "publicly"?

PR #1057, opened 2015-02-21
Make dictclient.py unicode capable

kyrias:


ProgVal:
So that's it? I only would have to do that to make it work? o_o

kyrias:
Yep \o/

Figured it out by trying the dictclient library manually and getting an encoding error.

PR #1063, opened 2015-02-28
Channel: allow adding an auto-rejoin delay

### LINKED ISSUE #1011: Channel: autorejoin after kick happens immediately
`supybot.plugins.Channel.alwaysRejoin` defaults to True and when the bot is kicked, it returns immediately which is bad behaviour even for bots.

The ZNC module `kickrejoin` has configurable rejoin period which defaults to 10 seconds and I think that feature of Channel should have something similar.

Sadly there are networks where there is no `/remove`...

GLolol:
This is done via the config variable `supybot.plugins.Channel.rejoinDelay`. Closes #1011.

The timing of the rejoins isn't super-exact because of `schedule.addEvent()` being used (the bot might join one second slower than what's set), but it should be fine for most purposes. When `rejoinDelay` is turned off, everything skips the scheduling and the rejoin is actually instant.

### DIFF
--- a/plugins/Channel/config.py
+++ b/plugins/Channel/config.py
@@ -51,6 +51,9 @@ conf.registerChannelValue(Channel, 'nicksInPrivate',
     registry.Boolean(True, _("""Determines whether the output of 'nicks' will
     be sent in private. This prevents mass-highlights of a channel's users,
     accidental or on purpose.""")))
-
+conf.registerChannelValue(Channel, 'rejoinDelay',
+    registry.NonNegativeInteger(0, _("""Determines how many seconds the bot will wait
+    before rejoining a channel if kicked and
+    supybot.plugins.Channel.alwaysRejoin is on.""")))
 
 # vim:set shiftwidth=4 softtabstop=4 expandtab textwidth=79:
--- a/plugins/Channel/plugin.py
+++ b/plugins/Channel/plugin.py
@@ -30,6 +30,7 @@
 
 import sys
 import fnmatch
+import time
 
 import supybot.conf as conf
 import supybot.ircdb as ircdb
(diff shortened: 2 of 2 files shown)


GLolol:
Test fail is irrelevant; something made `supybot-test` [segfault on pypy](https://travis-ci.org/ProgVal/Limnoria/jobs/52506435#L3927)?

plugins/Channel/config.py
```diff
@@ -51,6 +51,9 @@ def configure(advanced):
     registry.Boolean(True, _("""Determines whether the output of 'nicks' will
     be sent in private. This prevents mass-highlights of a channel's users,
     accidental or on purpose.""")))
-
+conf.registerChannelValue(Channel, 'rejoinDelay',
+    registry.NonNegativeInteger(0, _("""Determines how many seconds the bot will wait
```
Mikaela: I am not sure 0 is good default for this. WeeChat uses 30 for example and it's enough long time that people could think that the user could have written `/join #channel` by themselves and with the bot it could be thought that the bot admin gave the join command again.

ProgVal:
Could you add a name for the event?

Mikaela:
WeeChat calls the options as `irc.server_default.autorejoin` and `irc.server_default.autorejoin_delay`

plugins/Channel/config.py
GLolol: I kept it at 0 to stay consistent with current defaults.

Mikaela: I see

GLolol:
I looked around; other instances of `schedule.addEvent()` don't have names, and it seems that such a name must be unique for each event. Maybe something like `autoRejoin%s@%s % (channel, irc.network)` would do?


...And there's another merge conflict :(

ProgVal:
Hmm, indeed

PR #1066, opened 2015-02-28
Misc: give better help when command isn't found

### LINKED ISSUE #912: Misc: give better help when command isn't found
Instead of saying `Error: There is no command "blah".` you could say `Error: There is no command "blah". In case "blah" is a plugin, try "list blah".`

Line 301 in plugin.py of Misc.

GLolol:
Closes #912.

It basically looks like this:

```
<GLolol> asfadfa
-- Atlas: Error: "asfadfa" is not a valid command.
<GLolol> math
-- Atlas: Error: "math" is not a valid command. However, "Math" is the name of a loaded plugin, and you may be able to find its provided commands using 'list Math'.
```

### DIFF
--- a/plugins/Misc/plugin.py
+++ b/plugins/Misc/plugin.py
@@ -160,12 +160,20 @@ class Misc(callbacks.Plugin):
                                      'no command named %q in it.  Try "list '
                                      '%s" to see the commands in the %q '
                                      'plugin.'), plugin, tokens[1],
-                                     plugin, plugin))
+                                     plugin, plugin), Raise=True)
                 else:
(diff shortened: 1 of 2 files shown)


ProgVal:
Very nice! But could you fix the test case?

GLolol:
...I'm not quite sure what's causing the test fail.

```
<GLolol> asdfjkl
 -- Lily: Error: "asdfjkl" is not a valid command.
```

`Error: "asdfjkl" is not a valid command.` should match `not a valid command`, right?

ProgVal:
"timed out" means there was no reply.
Often caused by an exception being raised.

PR #1090, opened 2015-04-05
scripts/supybot: round total CPU time to 2 decimal places

GLolol:
This is a stylistic change, which prevents things such as 0.9400000000001 seconds from showing up.


ProgVal:
I think it would be more relevant to use `%.2f` instead (it keeps significant digits)

GLolol:
Whoops, I forgot about this. Updated.

Issue #1088, opened 2015-03-30
Testingn should should import from internal mock if available, otherwise, external mock


NP-Hardass:
Python 3 comes with an internal mock module, so ideally, the test system should import that if the external is not available.  Not sure if it is better to prioritize internal or external.

Reference, Python 3.4, no external mock package:
```
test_default_no_selected (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_default_yes_selected (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_no_selected_with_default_no (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_no_selected_with_default_yes (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_no_selected_without_defaults (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_yes_selected_with_default_no (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_yes_selected_with_default_yes (test_yn.TestYn) ... skipped 'python-mock is not installed.'
test_yes_selected_without_defaults (test_yn.TestYn) ... skipped 'python-mock is not installed.'
```
In comparison,  I have external mock installed for python 3.3, and those tests succeeded.

PR #1098, opened 2015-04-27
tests: Use internal mock from Python 3.3+ if available

GLolol:
Closes ProgVal/Limnoria#1088.


ProgVal:
I don't think it will work. If the import in the first except: block fails, it won't go to the second block.

GLolol:
Rebased.

PR #1121, opened 2015-06-08
ShrinkUrl plugin, add BitLy support.

t3chguy:
Login+ApiKey required for it to function.


nyuszika7h:
@t3chguy You forgot to update the translation files, but it was reverted anyway, so it doesn't matter.

PR #1131, opened 2015-06-28
Allow unregistered users to use 'add', 'remove', etc. in ChannelIdDatabasePlugin plugins

GLolol:
This adds a new configuration variable, `supybot.databases.plugins.requireRegistration`, which defaults to `True` for maximum security. When this is disabled, unauthenticated users will be able to use commands such as `quote add` freely. This is really my rationale for this, as most IRC quote bots I know of don't require registration, and people didn't like how Supybot forced registration for adding quotes.

Internally, this allows storing a string (hostmask) where the user ID (integer) usually goes, and `checkChangeAllowed` will `return True` if the hostmask of the caller matches the hostmask stored.

I also fixed Praise's tests by making them channel specific. (Travis CI pointed this out during building)

### DIFF
--- a/plugins/Quote/test.py
+++ b/plugins/Quote/test.py
@@ -40,4 +40,16 @@ class QuoteTestCase(ChannelPluginTestCase):
         self.assertRegexp("quote get 1", "goodbye")
         self.assertError("quote replace 5 afsdafas") # non-existant
 
+    def testUnauthenticatedAdd(self):
+        # This should fail because the user isn't registered
+        self.assertError('quote add hello world!')
+        original_value = conf.supybot.databases.plugins.requireRegistration()
+        conf.supybot.databases.plugins.requireRegistration.setValue(False)
+        try:
+            self.assertNotError('quote add hello world!')
+            self.assertRegexp('quote get 1', 'hello')
+            self.assertNotError('quote remove 1')
+        finally:
+            conf.supybot.databases.plugins.requireRegistration.setValue(original_value)
+
 # vim:set shiftwidth=4 softtabstop=4 expandtab textwidth=79:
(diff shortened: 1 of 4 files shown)


plugins/Quote/test.py
```diff
@@ -40,4 +40,16 @@ def testReplace(self):
...
+        self.assertError('quote add hello world!')
+        original_value = conf.supybot.databases.plugins.requireRegistration()
+        conf.supybot.databases.plugins.requireRegistration.setValue(False)
+        try:
+            self.assertNotError('quote add hello world!')
+            self.assertRegexp('quote get 1', 'hello')
+            self.assertNotError('quote remove 1')
+        finally:
+            conf.supybot.databases.plugins.requireRegistration.setValue(original_value)
+
```
ProgVal: You can do this instead:

```
        with conf.supybot.databases.plugins.requireRegistration.content(False):
            self.assertNotError('quote add hello world!')
            self.assertRegexp('quote get 1', 'hello')
            self.assertNotError('quote remove 1')
```

plugins/__init__.py
```diff
@@ -330,7 +330,7 @@ def noSuchRecord(self, irc, channel, id):
                   (self.name(), id, channel))
 
     def checkChangeAllowed(self, irc, msg, channel, user, record):
-        if user.id == record.by:
+        if (hasattr(user, 'id') and user.id == record.by) or user == record.by:
```
ProgVal: A comment explaining the logic of this condition would be nice

src/conf.py
```diff
@@ -931,6 +931,11 @@ def hasLinkChannel(channel):
         return lchannel
 
 registerGroup(supybot.databases, 'plugins')
+
+registerGlobalValue(supybot.databases.plugins, 'requireRegistration',
```
ProgVal: Can you make it channel-specific?

ProgVal:
Nice one, will merge it as soon as you reviewed my comments

GLolol:
I have to run out for a bit, I'll fix it when I get home. :)

PR #1132, opened 2015-07-03
Restore Google translate

Ban3:
The old translate function restored with a new working URL.


ProgVal:
Can you restore tests too?

GLolol:
Were there ever any proper tests for `translate`? e937128c69712d50b223b80dbb98a36ff5bb0470 seems to be the last commit before `translate` was removed, and faac1b0f2ba5294432fce64d89970a97863a2211 never touched any tests. Results might change over time as Google updates and improves its results too.

Ban3:
5b432a7 removed the test I restored. It was already removed in 2013, but I went ahead and put it back anyway. It was the only one I saw with a quick glance.

PR #1148, opened 2015-08-07
ircutils.formatWhois: ignore channels that have +p set

### LINKED ISSUE #98: Optionally hide channels the bot is in from 'network whois'
Issue sent from #limnoria at freenode by nyuszika7h (registered as nyuszika7h)

GLolol:
Fixes #98: see https://github.com/ProgVal/Limnoria/issues/98#issuecomment-121519435. The case of hiding channels for umode `+i` *isn't standard across IRCds*, so I don't think implementing it would be correct.


ProgVal:
What about `{'s', 'p'} & chanState.modes` instead?

GLolol:
Done, though that seems just as complex because `chanState.modes` is a dict, and not a set by default.

PR #1154, opened 2015-08-17
httpserver.py: Pass Content-Length to read call

mlq:
If a POST request is sent to the built-in http server the handling function does not terminate because the rfile.read() function blocks.

This patch passes the Content-Length value to the self.rfile.read() function that is required for it to do not block the method.

Regarding [RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4) the Content-Length header is expected to be sent otherwise this patch assumes a zero length.

### DIFF
--- a/src/httpserver.py
+++ b/src/httpserver.py
@@ -246,7 +246,8 @@ class SupyHTTPRequestHandler(BaseHTTPRequestHandler):
                          'CONTENT_TYPE':self.headers['Content-Type'],
                          })
         else:
-            form = self.rfile.read()
+            content_length = int(self.headers.get('Content-Length', '0'))
+            form = self.rfile.read(content_length)
         self.do_X('doPost', form=form)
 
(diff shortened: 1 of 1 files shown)


src/httpserver.py
```diff
@@ -246,7 +246,8 @@ def do_POST(self):
                          'CONTENT_TYPE':self.headers['Content-Type'],
                          })
         else:
-            form = self.rfile.read()
+            content_length = int(self.headers['Content-Length']) if 'Content-Length' in self.headers else 0
```
ProgVal: Do `content_length = int(self.headers.get('Content-Length', '0'))` here, it's shorter.

ProgVal:
thanks!

Pull requests go to the `testing` branch, not `master`, so I merged the branch manually: f45e2904f5a4eefe8b23e25685166d50e5106add

PR #1157, opened 2015-08-22
Alias improvements

jacksonmj:
711572b and fd53d8c are bugfixes.
54e9b4b - new feature, Alias.list command to list alias names, optionally filtered by --locked or --unlocked.
963a17c - new config option to allow relaxing the rules on alias names
d9b21dc (optional) - change default value for the new config option to a more relaxed one

I'm not sure what, if anything, needs doing about internationalisation and messages.pot though...

### DIFF
--- a/plugins/Alias/config.py
+++ b/plugins/Alias/config.py
@@ -44,6 +44,6 @@ Alias = conf.registerPlugin('Alias')
 conf.registerGroup(Alias, 'aliases')
 conf.registerGroup(Alias, 'escapedaliases')
 conf.registerGlobalValue(Alias, 'validName',
-    registry.String(r'^[a-z.|!?][a-z0-9.|!]*$', _("""Regex which alias names must match in order to be valid""")))
+    registry.String(r'^[^\x00-\x20]+$', _("""Regex which alias names must match in order to be valid""")))
 
 # vim:set shiftwidth=4 softtabstop=4 expandtab textwidth=79:
--- a/plugins/Alias/config.py
+++ b/plugins/Alias/config.py
@@ -43,5 +43,7 @@ def configure(advanced):
 Alias = conf.registerPlugin('Alias')
 conf.registerGroup(Alias, 'aliases')
 conf.registerGroup(Alias, 'escapedaliases')
+conf.registerGlobalValue(Alias, 'validName',
+    registry.String(r'^[a-z.|!?][a-z0-9.|!]*$', _("""Regex which alias names must match in order to be valid""")))
 
 # vim:set shiftwidth=4 softtabstop=4 expandtab textwidth=79:
--- a/plugins/Alias/plugin.py
+++ b/plugins/Alias/plugin.py
@@ -336,13 +336,9 @@ class Alias(callbacks.Plugin):
             irc.error(_('There is no such alias.'))
     unlock = wrap(unlock, [('checkCapability', 'admin'), 'commandName'])
 
-    _validNameRe = re.compile(r'^[a-z.|!?][a-z0-9.|!]*$')
     def addAlias(self, irc, name, alias, lock=False):
-        if not self._validNameRe.search(name):
-            raise AliasError('Names can only contain alphanumerical '
-                    'characters, dots, pipes, and '
-                    'exclamation/interrogatin marks '
-                    '(and the first character cannot be a number).')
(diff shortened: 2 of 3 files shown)


plugins/Alias/plugin.py
```diff
@@ -303,7 +306,12 @@ def lock(self, irc, msg, args, name):
         """
         if name in self.aliases and self.isCommandMethod(name):
             self.aliases[name][1] = True
-            conf.supybot.plugins.Alias.aliases.get(name).locked.setValue(True)
+            if (needsEscaping(name)):
+                group = conf.supybot.plugins.Alias.escapedaliases
+                name = escapeAlias(name)
+            else:
+                group = conf.supybot.plugins.Alias.aliases
```
ProgVal: Could you write a function for these 4 lines?

plugins/Alias/plugin.py
```diff
@@ -317,19 +325,20 @@ def unlock(self, irc, msg, args, name):
         """
         if name in self.aliases and self.isCommandMethod(name):
             self.aliases[name][1] = False
-            conf.supybot.plugins.Alias.aliases.get(name).locked.setValue(False)
+            if (needsEscaping(name)):
+                group = conf.supybot.plugins.Alias.escapedaliases
+                name = escapeAlias(name)
+            else:
+                group = conf.supybot.plugins.Alias.aliases
```
ProgVal: Same comment as above

plugins/Alias/plugin.py
```diff
@@ -412,6 +421,36 @@ def remove(self, irc, msg, args, name):
             irc.error(str(e))
     remove = wrap(remove, ['commandName'])
 
+    @internationalizeDocstring
+    def list(self, irc, msg, args, optlist):
+        """[--{locked,unlocked}]
```
ProgVal: I suggest `[--locked|--unlocked]` instead

plugins/Alias/config.py
```diff
@@ -43,5 +43,7 @@ def configure(advanced):
 Alias = conf.registerPlugin('Alias')
 conf.registerGroup(Alias, 'aliases')
 conf.registerGroup(Alias, 'escapedaliases')
+conf.registerGlobalValue(Alias, 'validName',
+    registry.String(r'^[^\x00-\x20]+$', _("""Regex which alias names must match in order to be valid""")))
```
ProgVal: You should make it more restrictive by using a whitelist instead of a blacklist.

Also, making this configurable makes it very easy for bot owners to break their bot + config file.

ProgVal:
Thanks for the PR. Please [fix the tests](https://travis-ci.org/ProgVal/Limnoria/jobs/76758118#L3627) and take into account my comments (either fix them or reply) and I will merge.

By the way, if you are interested in allowing more names, you can check out the Aka plugin, which is a rewrite of Alias with more flexible names.

plugins/Alias/config.py
jacksonmj: Pretty much all characters seem to work fine, with the exception of whitespace (disallowed by registry), and characters stripped by callbacks.canonicalName. The only reason I can see to exclude characters other than whitespace and callbacks.canonicalName stripped characters is that the bot owner does not want them in alias names. I'll add an additional non-configurable check for whitespace.
(You could keep it as a whitelist by not merging d9b21dc, which was deliberately placed in a separate commit).

I'm not sure how making this configurable allows bot owners to break their bot + config file. Can you provide an example?

ProgVal: If they set it to `^.*$`, then users can write arbitrary strings in the config file, right?

jacksonmj:
I think the solution to this is going to be "remove failing tests"...

ProgVal:
https://github.com/ProgVal/Limnoria/blob/testing/plugins/Aka/test.py#L151

plugins/Alias/config.py
jacksonmj: Strings written in the config file are restricted by supybot.registry, so:
":", ".", and "\" are escaped ("." is escaped by Alias too).
Alias names cannot contain whitespace (which includes newlines), since len(name.split()) must equal 1 otherwise the registry will refuse to accept the name.

jacksonmj:
Yay, that works. Thanks.

PR #1161, opened 2015-08-30
Send User.list response in private

jacksonmj:
Controlled by new config value plugins.User.listInPrivate.

This avoids mass-highlights of people who use their IRC nick as their bot username.

### DIFF
--- a/plugins/User/config.py
+++ b/plugins/User/config.py
@@ -43,7 +43,7 @@ def configure(advanced):
 
 
 User = conf.registerPlugin('User')
-conf.registerGlobalValue(User, 'listInPrivate',
+conf.registerChannelValue(User, 'listInPrivate',
     registry.Boolean(True, _("""Determines whether the output of 'user list'
     will be sent in private. This prevents mass-highlights of people who use
     their nick as their bot username.""")))
--- a/plugins/User/config.py
+++ b/plugins/User/config.py
@@ -43,8 +43,9 @@ def configure(advanced):
 
 
 User = conf.registerPlugin('User')
-# This is where your configuration variables (if any) should go.  For example:
-# conf.registerGlobalValue(User, 'someConfigVariableName',
-#     registry.Boolean(False, """Help for someConfigVariableName."""))
+conf.registerGlobalValue(User, 'listInPrivate',
+    registry.Boolean(True, _("""Determines whether the output of 'user list'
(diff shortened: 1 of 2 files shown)


plugins/User/config.py
```diff
@@ -43,8 +43,9 @@ def configure(advanced):
 
 
 User = conf.registerPlugin('User')
-# This is where your configuration variables (if any) should go.  For example:
-# conf.registerGlobalValue(User, 'someConfigVariableName',
-#     registry.Boolean(False, """Help for someConfigVariableName."""))
+conf.registerGlobalValue(User, 'listInPrivate',
```
ProgVal: Can you make it channel-specific?

plugins/User/config.py
```diff
@@ -43,8 +43,9 @@ def configure(advanced):
...
+    registry.Boolean(True, _("""Determines whether the output of 'user list'
```
ProgVal: I would prefer the default to be False, so it follows the generic behavior by default

plugins/User/config.py
```diff
@@ -43,8 +43,9 @@ def configure(advanced):
...
+    will be sent in private. This prevents mass-highlights of people who use
+    their nick as their bot username.""")))
```
ProgVal: You could add “This has no effect if supybot.reply.inPrivate is true”

plugins/User/config.py
GLolol: #1042 is similar and already defaults to True.

ProgVal: good point

plugins/User/plugin.py
```diff
@@ -88,7 +88,8 @@ def p(u):
                 users.append(u.name)
         if users:
             utils.sortBy(str.lower, users)
-            irc.reply(format('%L', users))
+            private = self.registryValue("listInPrivate")
```
ProgVal: `self.registryValue("listInPrivate", channel)` to make it channel-specific

plugins/User/config.py
jacksonmj: The config value appears to override supybot.reply.inPrivate. 
https://github.com/ProgVal/Limnoria/blob/testing/src/callbacks.py#L174-L175 - supybot.reply.inPrivate only takes effect if the private argument is None, and is ignored if private is True or False.

PR #1167, opened 2015-09-10
Filter: strip spaces in "unbinary" (Closes #1166)

### LINKED ISSUE #1166: Filter/unbinary: if the input has spaces, attempt stripping them and unbinarizing
Mostly all binary strings that I encounter have spaces in it and my bot replies to them `Error: '...' is not a valid binary string.`

However when I manually remove spaces it can unbinary the string and I think it could as well try to remove the spaces by itself.

Separate issue, but binary command could also use spaces (I am not sure where) to handle longer messages better.

* Supybot 0.83.4.1+limnoria 2015.03.17 running on Python3.2.3 (default, Feb 20 2013, 14:44:27) 
    * [GCC 4.7.2]

* Debian GNU/Linux 7.9 (wheezy)

GLolol:
This makes `_squish()` a shared function to prevent code duplication. Also, remove duplicate a `testUnbinary` function, and add tests for this.

### DIFF
--- a/plugins/Filter/plugin.py
+++ b/plugins/Filter/plugin.py
@@ -119,7 +119,7 @@ class Filter(callbacks.Plugin):
     hebrew = wrap(hebrew, ['text'])
 
     def _squish(self, text):
-        return ''.join(text.split())
+        return text.replace(' ', '')
 
     @internationalizeDocstring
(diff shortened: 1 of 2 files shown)


ProgVal:
Why not `.replace(' ', '')`? It's shorter and faster.

PR #1175, opened 2015-10-22
Web & utils.web: Fix for HTMLParser in Python 3.5.0

Ban3:
[Relevant Python issue.](http://bugs.python.org/issue23144)

With Python 3.5.0, Web returns an empty string as the title. The 'proper' solution to this is to force processing  by calling HTMLParser.close().

The fixed test case was failing on my machine because the actual title is capitalized.


ProgVal:
:+1:

PR #1177, opened 2015-10-25
Fixing fd.geturl() failing call in titleSnarfer

P-EB:
I met an issue in titleSnarfer call in Web plugin.

Here is my patch.


ProgVal:
Thanks for the PR, but PRs should go to the `testing`.
I fixed the problem in a way I think better: now getTitle() returns the target URL and the title, which avoids doing a second request like you did.

P-EB:
That seems good, thanks for taking this into account. I'll PR on testing next time. :)

PR #1186, opened 2015-11-10
Unix: ping 10 times instead of 5

Mikaela:
I believe 10 gives better picture on connectivity to the host than 5 and
is also more clear by 1 ping being 10%.


ProgVal:
I agree with #1187 being the right way

PR #1220, opened 2016-02-24
Socket: lower SSL related log level, and reword text to flow better

GLolol:
I don't like the idea of using critical-level logging to nag people to switch to TLS. Encouraging encryption is a good cause, but it's not a *critical software error* and shouldn't be treated as such.


ProgVal:
The motivation was that a security issue can be worse than a bug

GLolol:
It's more like the protocol's old and insecure by design, in which case it's not a security issue in the bot.

PR #1227, opened 2016-03-12
Network: fix getopts key for --nossl

GLolol:
c3dd5f8b64618de70414024c354b4867aaaa044c made plain text networks impossible to add. The getopts key needed to be changed from `ssl` to `nossl` in order to fix this.


ProgVal:
Oops.

Issue #1234, opened 2016-04-21
RSS plugin does not use the (HTTP) proxy configuration option


JunaidLoonat:
The `proxy` configuration option has the description of "Determines what proxy all HTTP requests should go through".

However, even if correctly configured, the [RSS plugin](https://github.com/ProgVal/Limnoria/tree/master/plugins/RSS) does not make use of any the HTTP proxy as one would expect. Instead, direct connections are made to retrieve the feeds for parsing.

ProgVal:
Could you test if that patch works?

```diff
diff --git a/plugins/RSS/plugin.py b/plugins/RSS/plugin.py
index f1a7bbc..1820e0c 100644
--- a/plugins/RSS/plugin.py
+++ b/plugins/RSS/plugin.py
@@ -52,6 +52,11 @@ import supybot.callbacks as callbacks
 from supybot.i18n import PluginInternationalization, internationalizeDocstring
 _ = PluginInternationalization('RSS')
 
+if minisix.PY2:
+    from urllib2 import ProxyHandler
+else:
+    from urllib.request import ProxyHandler
+
 def get_feedName(irc, msg, args, state):
     if ircutils.isChannel(args[0]):
         state.errorInvalid('feed name', args[0], 'must not be channel names.')
@@ -289,9 +294,13 @@ class RSS(callbacks.Plugin):
     # Feed fetching
 
     def update_feed(self, feed):
+        handlers = []
+        if utils.web.proxy():
+            handlers.append(ProxyHandler(
+                {'http': utils.force(utils.web.proxy())}))
         with feed.lock:
             d = feedparser.parse(feed.url, etag=feed.etag,
-                    modified=feed.modified)
+                    modified=feed.modified, handlers=handlers)
             if 'status' not in d or d.status != 304: # Not modified
                 if 'etag' in d:
                     feed.etag = d.etag
```

JunaidLoonat:
Thanks, the patch worked for Python 2(.7.6) but not for Python 3(.4.3)
I'll take a stab at (re-)wiring the RSS plugin back to util-web so there's no proxy configuring occurring within a plugin.

In addition, I'll try to push through other internal changes specifically with regards to HTTPS proxy usage and util-web improvements.

ProgVal:
On 25/04/2016 07:53, Junaid Loonat wrote:
> Thanks, the patch worked for Python 2(.7.6) but not for Python 3(.4.3)

Does it error, or just ignores the proxy setting?

[...]

PR #1235, opened 2016-04-25
Use HTTP proxy for HTTPS requests as well

JunaidLoonat:
Use proxy handler/opener classes, instead of request.set_proxy, to avoid any potential [bugs](http://bugs.python.org/issue1424152#msg68861) in older Python versions.

If configured, use the HTTP proxy for HTTPS requests as well.
_(Tested fine on both Python 2.7.6 and 3.4.3)_

This commit is a prerequisite for future commits related to #1234.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -1190,10 +1190,10 @@ class HttpProxy(registry.String):
                 'http': v,
                 'https': v
                 }
-    proxyHandler = ProxyHandler(proxies)
-    proxyOpenerDirector = build_opener(proxyHandler)
-    install_opener(proxyOpenerDirector)
-    super(HttpProxy, self).setValue(v)
+        proxyHandler = ProxyHandler(proxies)
+        proxyOpenerDirector = build_opener(proxyHandler)
+        install_opener(proxyOpenerDirector)
+        super(HttpProxy, self).setValue(v)
 
 registerGlobalValue(supybot.protocols.http, 'proxy',
     HttpProxy('', _("""Determines what proxy all HTTP requests should go
--- a/src/conf.py
+++ b/src/conf.py
@@ -34,6 +34,7 @@ import time
 import socket
 
 from . import ircutils, registry, utils
+from .utils import minisix
 from .version import version
 from .i18n import PluginInternationalization
 _ = PluginInternationalization()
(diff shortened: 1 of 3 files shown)


src/utils/web.py
```diff
@@ -145,7 +147,12 @@ def getUrlFd(url, headers=None, data=None, timeout=None):
             request.add_data(data)
         httpProxy = force(proxy)
         if httpProxy:
-            request.set_proxy(httpProxy, 'http')
+            proxyHandler = ProxyHandler({
+                'http': httpProxy,
+                'https': httpProxy
+                })
+            proxyOpener = build_opener(proxyHandler)
+            install_opener(proxyOpener)
```
ProgVal: That installs the proxy globally, until the next call to `getUrlFd`.
So if the config value is changed, this is not applied to plugins not using `getUrlFd`.

Setting the proxy from the config itself would be better.

JunaidLoonat: I'm just trying to understand your recommendation.

> That installs the proxy globally, until the next call to `getUrlFd`.

Yes. This shouldn't cause any issues since util-web is basically one of only two script files that makes direct use of `urllib`. The other file is a test script.

With `getUrlFd` being a relatively low-level method, it's called by all other request-related methods within util-web.
i.e. Any request-related method call would indirectly lead to the execution of `getUrlFd`, and thus the setting of the global proxy

> So if the config value is changed, this is not applied to plugins not using `getUrlFd`.

Correct, which is the same way it presently works without this commit.
Any plugin that does not currently make use of util-web for HTTP(S) requests will not benefit from this commit.

I believe we both agree that we should probably avoid the need for any plugins to unnecessarily configure the proxy themselves. Surely, all plugins should rather rely on the internal util-web component.

> Setting the proxy from the config itself would be better.

Just to confirm, is your main concern simply that the activation of a global HTTP(S) proxy should rather be done in within some configuration-related logic so that global proxy changes reflect immediately?
i.e. Update the global proxy as soon as the configuration option is changed

ProgVal: > Correct, which is the same way it presently works without this commit.
> Any plugin that does not currently make use of util-web for HTTP(S) requests will not benefit from this commit.

I was thinking about changing the proxy config.

Oh and by the way, I just noticed your code makes it impossible for users to disable the proxy.

> Just to confirm, is your main concern simply that the activation of a global HTTP(S) proxy should rather be done in within some configuration-related logic so that global proxy changes reflect immediately?

Yes. You can do that by adding a config type (ie. class) and overriding `set` or `setValue` (I never remember which one)

src/conf.py
```diff
@@ -1169,8 +1170,33 @@ def makeBanmask(self, hostmask, options=None, channel=None):
     similar.  It'll give up after it reads this many bytes, even if it hasn't
     found what it was looking for.""")))
 
+class HttpProxy(registry.String):
+  """Value must be a valid hostname:port string."""
+  def setValue(self, v):
+    if minisix.PY2:
+      from urllib2 import build_opener, install_opener, ProxyHandler
+    else:
+      from urllib.request import build_opener, install_opener, ProxyHandler
```
ProgVal: Please do these imports at the beginning of the module

src/conf.py
```diff
@@ -1169,8 +1170,33 @@ def makeBanmask(self, hostmask, options=None, channel=None):
...
+    proxies = {}
+    if v != "":
+      # TODO: improve checks
+      if ':' not in v:
+        self.error()
+      try:
+        int(v.rsplit(':', 1)[1])
+      except ValueError:
+        self.error()
```
ProgVal: There is probably a utility function somewhere (utils/ or urllib) to handle hostmask splitting.
Also, your code is broken if the user gives an IPv6 addresse without a port.

ProgVal: (nevermind that comment about IPv6)

JunaidLoonat:
Thanks @ProgVal for the advice and recommendations.
The change turned out much better within the config-related logic.

May I ask what's holding back the pull into testing?
The new code does the exact same validation as the existing SocksProxy class.
_(Shouldn't data validation changes/improvements should be done in a separate pull?)_

ProgVal:
> May I ask what's holding back the pull into testing?

This: `# TODO: improve checks`

> The new code does the exact same validation as the existing SocksProxy class.
> _(Shouldn't data validation changes/improvements should be done in a separate pull?)_

Hmm, if both have to be fixed, then yes.

JunaidLoonat:
@ProgVal, any comment on the validation routines? Are they fine?

PR #1239, opened 2016-05-05
Updated plugin to use Google CSE API instead of deprecated search api.

KaffeJunky123:
modified:   plugins/Google/config.py
	Added new configuration options for Google API Key and CSE ID.
modified:   plugins/Google/plugin.py
        Modified to conform with the google CSE API.
modified:   plugins/Google/test.py
	Modified to be useful for local testing.
	Set the API_KEY and CSE_ID envioronment varbiables for it to
	work.

### DIFF
--- a/plugins/Google/config.py
+++ b/plugins/Google/config.py
@@ -34,7 +34,7 @@ from supybot.i18n import PluginInternationalization, internationalizeDocstring
 _ = PluginInternationalization('Google')
 
 def configure(advanced):
-    from supybot.questions import output, yn
+    from supybot.questions import output, yn, something
     conf.registerPlugin('Google', True)
     output(_("""The Google plugin has the functionality to watch for URLs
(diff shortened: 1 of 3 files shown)


GLolol:
1) Pull requests go to the testing branch.
2) A fix for the Google plugin already exists in the testing branch: it uses HTML snarfing instead so an API key isn't required.

ProgVal:
Thanks for the PR.
However, I am closing it for the following reasons:
* PRs go to the testing branch
* I already fixed the issue (in testing) using HTML scraping
* I don't want Limnoria to depend on or promote third-party services, particularily commercial ones. The Google plugin exists for historical reasons, but I'm thinking about deprecating it.

ProgVal:
However, you can create your own plugin repo so PluginDownloader allows users to install your modified Google plugin easily

PR #1250, opened 2016-07-23
always sort the produced config files

mapreri:
I keep my stuff in a git repo, and I find pretty annoying to have to commit some imho nonsensical ordering change.

I'm not sure at all if this would break something else, I only quickly tested it and it seems to do what I'd expect.

I'm also unsure if it's preferred to have the main conf have a more meaningful order (more general/global options first, then the rest), but this would be a more bigger change (and not knowing supybot internals it would be impossible for me to do).

### DIFF
--- a/src/registry.py
+++ b/src/registry.py
@@ -192,7 +192,7 @@ def join(names):
 class Group(object):
     """A group; it doesn't hold a value unless handled by a subclass."""
     def __init__(self, help='', supplyDefault=False,
-                 orderAlphabetically=False, private=False):
+                 orderAlphabetically=True, private=False):
         self._help = utils.str.normalizeWhitespace(help)
         self._name = 'unset'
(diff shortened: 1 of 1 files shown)


ProgVal:
Did you check it does not significantly increase the flush time?

mapreri:
I can say that after the `flush` command over IRC the reply "The operation succeeded." came to me less than a second later, which seems to be the same as before.

Also consider that my instance where I tried this is not overly big, only being in 8 channels over 2 networks with a quite low number of plugins.

should I try to look for something else too?

ProgVal:
Merged manually in e41b75d636e1f9e2db5b6a92c8d16652d415f876 (the PR's target was master instead of testing)

mapreri:
yep, sorry, figured after I already opened it.

Thanks! :)

PR #1257, opened 2016-08-06
debian/compat is required, v9 or above

teward:
DebHelper, which helps build Debian packages, fails for the Debian packaging as is.  This is due to the system autoassuming a compat level of Debhelper 1, and anything older than v4 is not accepted.  Debian Testing, Ubuntu Xenial, etc. have DebHelper compat 9 as the recommended (this is also seen in the Debian Maintainer Guide (on August 6, 2016): https://www.debian.org/doc/manuals/maint-guide/dother.en.html#compat)

Failure to have this file results in Fail To Build Package with critical errors, in some build environments/setups (such as sbuild schroots).

This PR is for the Testing branch.


ProgVal:
@mapreri Your opinion on this?

teward:
Just a note: I tested builds back to ancient versions of debhelper; every old version of DebHelper will complain if the file doesn't exist, and doesn't have a value of 5 or greater in the file, as it will say "We don't support any debhelper versions older than 4".  5 adds in a margin of protection in this case.

ProgVal:
I see that @mapreri did the same thing in their repo, so I guess it's a good thing. Merging.

mapreri:
Yeah, d/compat is a file required by debhelper since somewhen last autumn.

Though, when bumping d/compat the version restriction of the debhelper build-dependency should be bumped too (to '(>= 9)', in this case; it doesn't particularly matter as version >=9 has been with us for the last several years, but is custom do to so anyway (and possibly more correct().

PR #1258, opened 2016-08-06
[master] debian/compat is required, v9 or above

teward:
DebHelper, which helps build Debian packages, fails for the Debian packaging as is.  This is due to the system autoassuming a compat level of Debhelper 1, and anything older than v4 is not accepted.  Debian Testing, Ubuntu Xenial, etc. have DebHelper compat 9 as the recommended (this is also seen in the Debian Maintainer Guide (on August 6, 2016): https://www.debian.org/doc/manuals/maint-guide/dother.en.html#compat)

Failure to have this file results in Fail To Build Package with critical errors, in some build environments/setups (such as sbuild schroots).

This PR is for the Master branch.  There is a separate PR for the Testing branch; if these don't work together, then please reject this one and accept Testing's PR instead.


ProgVal:
> This PR is for the Master branch. There is a separate PR for the Testing branch; if these don't work together, then please reject this one and accept Testing's PR instead.

Indeed, they don't. master always points to an ancestor of the commit pointed by testing.

teward:
A side note then: anyone building master will have to include a compat file of their own to get successful builds.  Debian debhelper is what triggered the error, even on DH version 7 (which is old)

PR #1261, opened 2016-08-19
strip duplicate nicks from @tell

cottongin:
Removes duplicate nicks from @tell <same nick>,<same nick>,<same nick> to avoid spam

### DIFF
--- a/plugins/Later/plugin.py
+++ b/plugins/Later/plugin.py
@@ -157,6 +157,9 @@ class Later(callbacks.Plugin):
         contain wildcard characters, and the first matching nick will be
         given the note.
         """
+        # strip duplicates
+        nicks = list(set(nicks))
+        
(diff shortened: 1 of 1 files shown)


ProgVal:
It won't prevent users from using `@tell foo,foo*,foo**`

cottongin:
No real way to fix that I can think of, without breaking wildcards completely.

cottongin:
Upon testing foo*, foo**, foo*** on my end, it appears to work as intended

ProgVal:
uh, actually, wildcards are not allowed, wtf

ProgVal:
Nevermind, it's because the only `*` allowed is the last character.

cottongin:
On further testing, since sets are unordered, it can actually break with `**` and `***`.  `*` works as intended.

PR #1262, opened 2016-08-19
Strip duplicate nicks given to @tell to avoid spam.

cottongin:
further improvment on #1261


ProgVal:
I believe this was handled by an other commit. Closing.

PR #1267, opened 2016-10-01
conf: fix default and help for supybot.user

GLolol:
- "Limnoria $version" would expand to "Limnoria Limnoria 1234.56.78", which is probably not what's intended
- Empty real names no longer expand to anything since e5729bc86d8148dcb39339d488b674e125fe0aa9


ProgVal:
> "Limnoria $version" would expand to "Limnoria Limnoria 1234.56.78", which is probably not what's intended

This was already fixed by an other commit.

> Empty real names no longer expand to anything since e5729bc

Reverted by 97d29d7

PR #1273, opened 2016-11-27
User: allow for custom whoami error (fixes #1142)

### LINKED ISSUE #1142: Allow customizing the "whoami" unidentified message
Like `help` can be customized.

> I don't recognize you. You can message me either of these two commands: "user identify <username> <password>" to log in or "user register <username> <password>" to register.

I personally don't give users passwords (`supybot-adduser -u someone -p $(pwgen -s 50) -c \#channel,op`) and just identify them with hostmask or NickServ account and also have disabled registration, so these instructions don't apply to my instance.

I would like to change it to e.g. "You aren't identified. If you think you should be, use the 'auth' command. If you aren't registered and fill the criteria of https://mikaela.info/bot#users, ask Mikaela."

IotaSpencer:
May need to fix some internationalization, but this takes care of a easy feature request.

### DIFF
--- a/plugins/User/plugin.py
+++ b/plugins/User/plugin.py
@@ -480,7 +480,7 @@ class User(callbacks.Plugin):
             user = ircdb.users.getUser(msg.prefix)
             irc.reply(user.name)
         except KeyError:
-            irc.reply(self.registryValue('customWhoamiError'))
+            irc.reply(self.registryValue('whoamiError'))
     whoami = wrap(whoami)
 
     @internationalizeDocstring
--- a/plugins/User/plugin.py
+++ b/plugins/User/plugin.py
@@ -480,7 +480,10 @@ class User(callbacks.Plugin):
             user = ircdb.users.getUser(msg.prefix)
             irc.reply(user.name)
         except KeyError:
-            irc.reply(_('I don\'t recognize you. You can message me either of these two commands: "user identify <username> <password>" to log in or "user register <username> <password>" to register.'))
+            if not self.registryValue('customWho'):
+                irc.reply(_('I don\'t recognize you. You can message me either of these two commands: "user identify <username> <password>" to log in or "user register <username> <password>" to register.'))
+            else:
+                irc.reply(self.registryValue('customWho'))
     whoami = wrap(whoami)
 
     @internationalizeDocstring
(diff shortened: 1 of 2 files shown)


ProgVal:
Why do you use `registry.NormalizedString` and not `registry.String`?

Also, instead of having None as a default and have the plugin handle None as a special case, you could put the default message as the default value of `customWho`.
This will fix the crash of test cases, and allow users to see the default string on IRC.

Also, could you rename `customWho` to `customWhoamiError`? I believe that makes its purpose clearer.

plugins/User/plugin.py
```diff
@@ -480,7 +480,7 @@ def whoami(self, irc, msg, args):
             user = ircdb.users.getUser(msg.prefix)
             irc.reply(user.name)
         except KeyError:
-            irc.reply(_('I don\'t recognize you. You can message me either of these two commands: "user identify <username> <password>" to log in or "user register <username> <password>" to register.'))
+            irc.reply(self.registryValue('customWho'))
```
GLolol: It has to be `'customWhoamiError'` here as well.

plugins/User/plugin.py
```diff
@@ -480,7 +480,7 @@ def whoami(self, irc, msg, args):
             user = ircdb.users.getUser(msg.prefix)
             irc.reply(user.name)
         except KeyError:
-            irc.reply(self.registryValue('customWho'))
+            irc.reply(self.registryValue('customWhoError'))
```
ProgVal: `customWhoamiError`, not `customWhoError`

IotaSpencer: Yes just fixed that, lot of stuff on my mind right now.

ProgVal:
One last thing: could you wrap the default value of the config var in `_()`?

IotaSpencer:
there should be all fixed

ProgVal:
Sorry for not seeing it earlier, but the content of the value is not normalized like the help string is, so you have to do it like this:

```
_("I don\'t recognize you. You can message me either of "
    "these two commands: \"user identify <username> <password>\" to log in or "
    "\"user register <username> <password>\" to register.")
```

nyuszika7h:
Nitpick, I'd make it just `whoamiError` because it's not "custom" if it's the default value.

ProgVal:
After giving it some thought, I prefer the original solution, because it makes it possible to alter the message in future versions for existing bots.
I'll cherry-pick your first commit and fix it and stop bothering you with my changes.

ProgVal:
b46ea6f6bced7e10298a6a24506db6e1baceb279

Issue #896, opened 2014-11-06
Fun & Games with OpenShift (AKA Making bot usable on PAAS)


Efreak:
I'm running supybot on openshift, but I've encountered a few problems that should actually be relatively simple to solve.

I'm modifying the sourcecode directly to use environment variables. Can you add an expander for `os.environ['some_var']`? The best way would probably be to expand _any_ environment variable. It should be available for file locations, ip address, port, etc.
I've removed references to addresses6 and replaced addresses4 with `addresses4 = [(4, (os.environ['OPENSHIFT_DIY_IP'], int(os.environ['OPENSHIFT_DIY_PORT'])))]`

Also, botchk should be able to either grab a tail of a log in some way other than using pipes, as pipes aren't allowed, or it should gracefully fail on that and just know that the bot's running but inaccessible.
`PermissionError: [Errno 13] Permission denied`

If supybot webserver logs incoming ip addresses, it should be able to parse the X-Forwarded-For and so forth (I haven't checked this, because I simply don't care).

This is all that's required for getting supybot running on heroku, openshift, etc. Supybot is difficult to set up on openshift's python cartridges, but I can provide a step-by-step of how I set it up on their DIY cartridge if you want.

rnc:
@Efreak @ProgVal I have a working installation now of Limnoria on OpenShift V3 (I have tried it running locally via minishift or a hosted cluster). It is using https://github.com/rnc/docker-limnoria which is derived from your https://github.com/ProgVal/docker-limnoria ; I am happy to send a PR for that although it is a bit different currently.

PR #1279, opened 2016-12-23
Use copyfile instead of copy

rnc:
The reason this would be useful is related to #896 . 

I have been trying to get Limnoria running on OpenShift v3 partly by using https://github.com/ProgVal/docker-limnoria (which I have some modifications for) and using https://github.com/minishift/minishift to run a local OpenShift cluster. 

According to https://docs.python.org/2/library/shutil.html copy attempts to set the ,permission bits while copyfile just copies the contents. This change should therefore have no side-affects in day to day use but makes Limnoria usable on OpenShift.

In OpenShift the user ID is random (and you are not running as root) therefore I was getting 
```[Errno 1] Operation not permitted: '/var/supybot/supybot.conf.bak```

### DIFF
--- a/scripts/supybot
+++ b/scripts/supybot
@@ -208,7 +208,7 @@ if __name__ == '__main__':
             # The registry *MUST* be opened before importing log or conf.
             i18n.getLocaleFromRegistryFilename(registryFilename)
             registry.open_registry(registryFilename)
-            shutil.copy(registryFilename, registryFilename + '.bak')
+            shutil.copyfile(registryFilename, registryFilename + '.bak')
         except registry.InvalidRegistryFile as e:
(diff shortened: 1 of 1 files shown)


rnc:
@ProgVal  When does testing branch get merged to master please?

ProgVal:
I would like to wait a week or two to see if anyone complains about this change.

ProgVal:
I would like to wait a week or two to see if anyone complains about this change.
Is it urgent I merge it?

rnc:
@ProgVal Its not urgent, no worries :-)
I do wonder if the other uses of copy should be migrated to copyfile although it appears in standard use, I have not hit any further issues in OpenShift so far.

PR #1280, opened 2017-01-21
Allow specifying the format of Owner.announce

nyuszika7h:
Someone asked how to remove the owner's username from announcements, so it may be useful to make this configurable instead of making them edit a core plugin's code which is not the best idea, or write another plugin.

### DIFF
--- a/plugins/Owner/config.py
+++ b/plugins/Owner/config.py
@@ -45,6 +45,11 @@ Owner = conf.registerPlugin('Owner', True)
 conf.registerGlobalValue(Owner, 'public',
     registry.Boolean(True, """Determines whether this plugin is publicly
     visible."""))
+conf.registerGlobalValue(Owner, 'announceFormat',
+    registry.String('Announcement from my owner ($owner): $message',
+    """Determines the format of messages sent by the @announce command.
+    $owner may be used for the username of the owner calling this command,
+    and $message for the announcement being made."""))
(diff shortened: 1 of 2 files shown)


ProgVal:
Could you use a template string instead of %-formatting? (for security reasons and consistancy with other config values)

nyuszika7h:
Sure. I would have went for `.format()`, but Limnoria seems to use `%`-formatting in most places, and I assumed `$` templates are just used for common things like `$nick`, `$channel` and such.

nyuszika7h:
Done.

ProgVal:
> common things like `$nick`, `$channel` and such

Hey, that's a good idea. Could you use ircutils.standardSubstitute so
these are available?

nyuszika7h:
Done, I changed `$user` to `$owner` to avoid a clash as there doesn't seem to be an existing variable for the actual username in the bot (maybe there should be).

nyuszika7h:
Wait, do not merge this yet, it's broken.

ProgVal:
> maybe there should be

Yes, maybe. But I am not sure how to make it behave if the user is not logged in.

ProgVal:
Instead of substituting twice, could you use the `env` parameter of `standardSubstitute`

ProgVal:
And `$user` is already used for the ident part of the prefix

nyuszika7h:
I didn't see your messages but I figured that out. Fixed.

PR #1284, opened 2017-02-24
ircutils: remove redundant str.replace calls

GLolol:
As strings, `'\x0f' == '\x0F'`, and `'\x1f' == '\x1F'`, meaning these extra replace calls are redundant.


GLolol:
The test fail seems to be unrelated: https://travis-ci.org/ProgVal/Limnoria/jobs/204820399

Issue #252, opened 2011-10-28
Network-based configuration variables.


ProgVal:

ProgVal:
I added a `network-config` branch to the repository, with an initial implementation of net+chan specific variables.

ProgVal:
It's merged and released now.

PR #1299, opened 2017-08-03
Make irc.state.history network specific

### LINKED ISSUE #1211: Should each IRC network object have its own, distinct irc.state.history?
I noticed that the `irc.state.history` attribute stores the command history of _all_ networks, instead of just the ones associated with that network. Since other `irc.state` variables (IRCd name, available modes, etc.) are all specific to each connection, it would make sense if `irc.state.history` was also connection-specific. Is this an oversight or simply a design decision I wasn't aware of?

The rationale for this is a sed-style replacer bot that runs on multiple networks, but in two different channels with the same name (e.g. `#chat`). The bot starts to confuse messages from the two channels, since messages from both are added to the same `irc.state.history`.

GLolol:
It is fairly counterintuitive for plugins to read the IrcState history, only to find that the message it got is from the wrong network.
I'm not sure why the original Supybot chose to link all the history buffers together, as those lines came from before Supybot was introduced to Git.

Closes #1211.

----

I realize that this is a breaking API change, which may subtly break any plugins relying on history to be global. I don't know of any specific plugins that would break however...

### DIFF
--- a/plugins/Owner/plugin.py
+++ b/plugins/Owner/plugin.py
@@ -160,9 +160,6 @@ class Owner(callbacks.Plugin):
                    'No servers are set for the %s network.' % network
         self.log.debug('Creating new Irc for %s.', network)
         newIrc = irclib.Irc(network)
-        for irc in world.ircs:
-            if irc != newIrc:
-                newIrc.state.history = irc.state.history
         driver = drivers.newDriver(newIrc)
         self._loadPlugins(newIrc)
         return newIrc


ProgVal:
I find these three lines strange. First, the `if` clause is not needed, as the `newIrc` is not yet added to `world.ircs`.
Also, the `history` could be made a parameter to `irclib.Irc`. 

So my guess is that was a quick hack; but I cannot see a reason for it.

I'll give this PR some more thought before merging.

jamessan:
c0386751cfb470e1dfd3925b5060c997c6b45f73 is the commit that introduced the for loop.

898b21987173a0e0b77a103826b6ef62665ff6f4 is where copying the history to newly connected Irc objects was added.

It was definitely intentional to share history across networks.  That's done to allow having a common history for channels that are being relayed between networks.

As far as the for loop and if, those are indeed completely useless.  `irclib.Irc` implements `__eq__` by comparing the `id()` of each object, so they'll never be the same here given that `newIrc` was just created.  It probably should have just been

```python
if world.ircs:
    newIrc.state.history = world.ircs[0].state.history
```

Now, I think Limnoria reimplemented the way relaying is handled, so there's fine-grained control over which networks relay which channels, and the channel names don't need to be equal on each network.  That's quite different what was being supported with this code.

Supybot assumed that if you `Relay.join`ed a channel, it would be joined on all networks.  Hmm, however we did allow you to `Network.join` on different networks, which therefore wouldn't be relaying.  In that case, there is potential leaking of information across networks.

I guess it just boils down to not having good separation of networks. :)

ProgVal:
@jamessan Thanks for this comment!

Limnoria uses the same logic as Supybot in the Relay plugin; but LinkRelay (a third-party plugin) is more popular and does what you describe.
However, it is true that Limnoria has more network-specific settings than supybot (nick, ident, username, umodes, ...), on (many) user requests.
There are also a few more things I would like to make network-specific because of user requests (#252, #1179, ...)

GLolol:
Unless I'm missing something, I don't see any mention of `state.history` in the [current incarnation of Relay](https://github.com/ProgVal/Limnoria/blob/master/plugins/Relay/plugin.py). So, is there still a backwards compat issue that needs to be worked around?

jamessan:
There isn't any use of `state.history` in the Relay plugin for Supybot either.  It's the other plugins, like [Misc.last](https://github.com/Supybot/Supybot/blob/master/plugins/Misc/plugin.py#L326), that utilized the history.  The intention being that `@last` on a Relayed channel would show the last message, independent of network.

However, as I alluded to earlier, the histories were shared regardless of whether relaying was happening, which is a potential leak of information.  The history should only be shared when the channels are shared, if at all.  Of course, then you would also need a way to disentangle the histories if the relaying stopped.

In summary, I think the original implementation favored the simplicity of sharing under the assumption that the same channels on different networks would likely be relayed.  In hindsight, it probably would have been better to keep the histories unique across networks but provided common ways to interweave the histories, when needed.

PR #1301, opened 2017-09-17
Add config variable supybot.commands.allowShell.

ProgVal:
This PR add a new configuration variable, named `supybot.commands.allowShell`, whose help:

> Allows this bot's owner user to use commands
> that grants them shell access. This commands exists in case you want to
> prevent MITM from the IRC network itself (vulnerable IRCd or IRCops)
> from gaining shell access to the bot's server by impersonating the owner.
> Setting this to False also disables plugins and commands that can be
> used to indirectly gain shell access.

When set to False, its effects are the following:

* Config plugin:
  * Prevent changes to `supybot.commands.allowShell`, so it cannot be re-enabled by an attacker
  * Prevent changes to variables in `supybot.directories.plugins`, so an attacker able to upload files on the server (eg. a web app / CMS) cannot load arbitrary code
  * Prevent changes to other variables in `supybot.directories`, so an attacker cannot write arbitrary files
  * Disable `@export` so an attacker cannot write arbitrary files
* Debug: disable most commands
* Karma: disable `@dump` and `@load`, to prevent writing and reading arbitrary files
* PluginDownloader: disable `@install` to prevent installing plugins from third-party repositories, which may be vulnerable (intentionally or not)
* Unix: disable `@call` and `@shell`

Note that when `supybot.commands.allowShell` is True it can be set to False via IRC, but if it is False, it may only be changed by editing the `.conf` file.

This variable defaults to True, which preserves the behavior of Limnoria before this PR.
The rational for this default is to have PluginDownloader usable by default, because installing plugins is hard for beginners.

This variable is *not* network-specific on purpose. There are many ways to run a command from a network in the context of another network (`@network.command`, Scheduler, ...), and I wouldn't trust myself to write code preventing all of these.

Any question or remark?

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -748,6 +748,16 @@ registerGlobalValue(supybot.commands.defaultPlugins, 'importantPlugins',
         know what you're doing, then also know that this set is
         case-sensitive.""")))
 
+# For this config variable to make sense, it must no be writable via IRC.
+# Make sure it is always blacklisted from the Config plugin.
+registerGlobalValue(supybot.commands, 'allowShell',
+    registry.Boolean(True, _("""Allows this bot's owner user to use commands
+    that grants them shell access. This commands exists in case you want to
+    prevent MITM from the IRC network itself (vulnerable IRCd or IRCops)
+    from gaining shell access to the bot's server by impersonating the owner.
+    Setting this to False also disables plugins and commands that can be
+    used to indirectly gain shell access.""")))
+
 # supybot.commands.disabled moved to callbacks for canonicalName.
 
 ###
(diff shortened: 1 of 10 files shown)


src/conf.py
```diff
@@ -748,6 +748,16 @@ class ValidBrackets(registry.OnlySomeStrings):
         know what you're doing, then also know that this set is
         case-sensitive.""")))
 
+# For this config variable to make sense, it must no be writable via IRC.
+# Make sure it is always blacklisted from the Config plugin.
+registerGlobalValue(supybot.commands, 'allowShell',
+    registry.Boolean(True, _("""Allows this bot's owner user to use commands
+    that grants them shell access. This commands exists in case you want to
```
GLolol: `s/This commands/This config option/` ?

ProgVal: Indeed. Fixed.

PR #1302, opened 2017-09-17
Substitute $nick and friends with <console> when exiting on Ctrl-C

GLolol:
This is probably better than leaving them as raw strings.

### DIFF
--- a/scripts/supybot
+++ b/scripts/supybot
@@ -117,7 +117,14 @@ def main():
                 for irc in world.ircs:
                     quitmsg = conf.supybot.plugins.Owner.quitMsg() or \
                               'Ctrl-C at console.'
-                    quitmsg = ircutils.standardSubstitute(irc, None, quitmsg)
+                    # Because we're quitting from the console, none of the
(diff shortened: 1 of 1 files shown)


scripts/supybot
```diff
@@ -117,7 +117,14 @@ def main():
                 for irc in world.ircs:
                     quitmsg = conf.supybot.plugins.Owner.quitMsg() or \
                               'Ctrl-C at console.'
-                    quitmsg = ircutils.standardSubstitute(irc, None, quitmsg)
+                    # Because we're quitting from the console, none of the
+                    # standard msg substitutions exist, and these will show as
+                    # raw strings by default. Substitute them here with
+                    # something meaningful instead.
+                    env = dict((key, '<console>') for key in ('who', 'nick',
+                                                              'user', 'host'))
```
ProgVal:
```
env = dict((key, '<console>')
           for key in ('who', 'nick', 'user', 'host'))
```

GLolol: Updated

Issue #686, opened 2014-05-27
Owner or Admin: add commands to perform command on all servers / channels.


Mikaela:
Some IRC clients have commands like `/allserv`, `/allchan` and `/allchanl`.
1. Executes command on all serveres.
2. Executes command on all channels on all servers.
3. Executes command on all channels on current server.

Oh, and I mean bot commands, not raw commands.

And if the command changes default configuration, it should be applied to global and channelspecific variables.

GLolol:
Fixed by #1305 (that PR isn't written to auto-close this issue)

PR #1305, opened 2017-09-22
Network: add all+server+channels and all+channels commands

IotaSpencer:
This would fulfill #686

### DIFF
--- a/plugins/Network/plugin.py
+++ b/plugins/Network/plugin.py
@@ -157,7 +157,7 @@ class Network(callbacks.Plugin):
     command = wrap(command, ['admin', ('networkIrc', True), many('something')])
     
     def cmdall(self, irc, msg, args, commandAndArgs):
-        """<command> <args>...
+        """<command> [<arg> ...]
         
         Perform <command> (with its associated <arg>s) on all networks.
         """
@@ -167,21 +167,23 @@ class Network(callbacks.Plugin):
     cmdall = wrap(cmdall, ['admin', many('something')])
 
     def cmdallchans(self, irc, msg, args, commandAndArgs):
-        """<command> <args>...
+        """<command> [<arg> ...]
+
         Perform <command> (with associated <arg>s) on all networks and all channels."""
         for ircd in world.ircs:
-            for channel in ircd.state.channels.keys():
+            for channel in ircd.state.channels:
                 msg.args[0] = channel
                 self.Proxy(ircd, msg, commandAndArgs)
     cmdallchans = wrap(cmdallchans, ['admin', many('something')])
 
     def acmd(self, irc, msg, args, commandAndArgs):
-        """<command> <args>...
+        """<command> [<arg> ...]
+
         Perform <command> (with associated <arg>s on all channels on current network."""
-        for channel in irc.state.channels.keys():
+        for channel in irc.state.channels:
             msg.args[0] = channel
             self.Proxy(irc, msg, commandAndArgs)
-    achanscmd = wrap(acmd, ['admin', many('something')])
+    acmd = wrap(acmd, ['admin', many('something')])
 
     ###
     # whois command-related stuff.
--- a/plugins/Network/plugin.py
+++ b/plugins/Network/plugin.py
@@ -165,7 +165,24 @@ class Network(callbacks.Plugin):
         for ircd in ircs:
             self.Proxy(ircd, msg, commandAndArgs)
     cmdall = wrap(cmdall, ['admin', many('something')])
-    
+
+    def asrvchanscmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s) on all networks and all channels."""
+        ircs = world.ircs
+        for ircd in ircs:
+            for channel in ircd.state.channels.keys():
+                self.Proxy(ircd, msg, commandAndArgs)
+    asrvchanscmd = wrap(asrvchanscmd, ['admin', many('something')])
+
+    def achanscmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s on all channels on current network."""
+        ircd = world.getIrc(irc.network)
+        for channel in irc.state.channels.keys():
+            self.Proxy(ircd, msg, commandAndArgs)
+    achanscmd = wrap(achanscmd, ['admin', many('something')])
+
     ###
     # whois command-related stuff.
     ###
--- a/plugins/Network/plugin.py
+++ b/plugins/Network/plugin.py
@@ -166,25 +166,6 @@ class Network(callbacks.Plugin):
             self.Proxy(ircd, msg, commandAndArgs)
     cmdall = wrap(cmdall, ['admin', many('something')])
 
-    def cmdallchans(self, irc, msg, args, commandAndArgs):
-        """<command> [<arg> ...]
-
-        Perform <command> (with associated <arg>s) on all networks and all channels."""
-        for ircd in world.ircs:
-            for channel in ircd.state.channels:
-                msg.args[0] = channel
-                self.Proxy(ircd, msg, commandAndArgs)
-    cmdallchans = wrap(cmdallchans, ['admin', many('something')])
-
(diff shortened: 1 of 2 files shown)


plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+            for channel in ircd.state.channels.keys():
+                self.Proxy(ircd, msg, commandAndArgs)
+    asrvchanscmd = wrap(asrvchanscmd, ['admin', many('something')])
+
+    def achanscmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s on all channels on current network."""
+        ircd = world.getIrc(irc.network)
+        for channel in irc.state.channels.keys():
+            self.Proxy(ircd, msg, commandAndArgs)
```
GLolol: This is misleading because doesn't actually change the target channel that the command is being run in. Suppose you're in `#test` and run an echo:

```
06:51 <~GLolol> `achanscmd echo $channel
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
06:51 <@Lily> ​#test
```

All that happens is that the same command runs multiple times, once for each channel the bot is in. Maybe you could try overriding `msg.args[0]`?

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+    def asrvchanscmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s) on all networks and all channels."""
+        ircs = world.ircs
+        for ircd in ircs:
+            for channel in ircd.state.channels.keys():
+                self.Proxy(ircd, msg, commandAndArgs)
+    asrvchanscmd = wrap(asrvchanscmd, ['admin', many('something')])
+
+    def achanscmd(self, irc, msg, args, commandAndArgs):
```
GLolol: I prefer a name like `cmdallchans` to be consistent with `cmdall`'s naming.

plugins/Network/plugin.py
GLolol: Also, you can just use `self.Proxy(irc, ...)` instead of `getIrc` on the current network.

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
         for ircd in ircs:
             self.Proxy(ircd, msg, commandAndArgs)
     cmdall = wrap(cmdall, ['admin', many('something')])
-    
+
+    def asrvchanscmd(self, irc, msg, args, commandAndArgs):
```
GLolol: Is this command really needed? You should be able to run `asrvchanscmd` through `cmdall`.

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+        """<command> <args>...
+        Perform <command> (with associated <arg>s) on all networks and all channels."""
+        ircs = world.ircs
+        for ircd in ircs:
```
GLolol: You can just iterate over `world.ircs`, no need for a separate variable.

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+            for channel in ircd.state.channels.keys():
+                self.Proxy(ircd, msg, commandAndArgs)
```
GLolol: Same problem as below, it doesn't change the channel the command is actually being run in.

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
         for ircd in ircs:
             self.Proxy(ircd, msg, commandAndArgs)
     cmdall = wrap(cmdall, ['admin', many('something')])
-    
+
+    def cmdallchans(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s) on all networks and all channels."""
```
ProgVal: add a newline between the syntax and the help

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+        for ircd in world.ircs:
+            for channel in ircd.state.channels.keys():
+                msg.args[0] = channel
+                self.Proxy(ircd, msg, commandAndArgs)
+    cmdallchans = wrap(cmdallchans, ['admin', many('something')])
+
+    def acmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s on all channels on current network."""
```
ProgVal: add a newline between the syntax and the help

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
         for ircd in ircs:
             self.Proxy(ircd, msg, commandAndArgs)
     cmdall = wrap(cmdall, ['admin', many('something')])
-    
+
+    def cmdallchans(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s) on all networks and all channels."""
+        for ircd in world.ircs:
+            for channel in ircd.state.channels.keys():
```
ProgVal: `for channel in ircd.state.channels:`

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+                msg.args[0] = channel
+                self.Proxy(ircd, msg, commandAndArgs)
+    cmdallchans = wrap(cmdallchans, ['admin', many('something')])
+
+    def acmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
+        Perform <command> (with associated <arg>s on all channels on current network."""
+        for channel in irc.state.channels.keys():
```
ProgVal: `for channel in irc.state.channels:`

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+            msg.args[0] = channel
+            self.Proxy(irc, msg, commandAndArgs)
+    achanscmd = wrap(acmd, ['admin', many('something')])
```
ProgVal: doesn't match the function name

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
         for ircd in ircs:
             self.Proxy(ircd, msg, commandAndArgs)
     cmdall = wrap(cmdall, ['admin', many('something')])
-    
+
+    def cmdallchans(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
```
ProgVal: `<command> [<arg> ...]`

plugins/Network/plugin.py
```diff
@@ -165,7 +165,24 @@ def cmdall(self, irc, msg, args, commandAndArgs):
...
+        Perform <command> (with associated <arg>s) on all networks and all channels."""
+        for ircd in world.ircs:
+            for channel in ircd.state.channels.keys():
+                msg.args[0] = channel
+                self.Proxy(ircd, msg, commandAndArgs)
+    cmdallchans = wrap(cmdallchans, ['admin', many('something')])
+
+    def acmd(self, irc, msg, args, commandAndArgs):
+        """<command> <args>...
```
ProgVal: `<command> [<arg> ...]`

ProgVal:
Actually, what about adding an option named `--all-networks` to `@network command`, instead of adding `@acmd`?

PR #1306, opened 2017-09-23
Unix: make the default ping count configurable (fixes #1187)

### LINKED ISSUE #1187: Unix: allow specifying the default ping count
I opened https://github.com/ProgVal/Limnoria/pull/1186 in case people agree with my preference, but it doesn't resolve the issue why I opened it.
- Debian GNU/Linux 7.9 (wheezy)
- The current (running) version of this Supybot is 0.83.4.1+limnoria 2015.10.04, running on Python 3.5.0 (default, Oct  4 2015, 09:45:03)  [GCC 4.2.1 Compatible Debian Clang 3.0 (tags/RELEASE_30/final)].  The newest versions available online are 2015.10.04 (in master), 2015.11.09 (in testing).

IotaSpencer:


ProgVal:
Doesn't this code raise a TypeError?

ProgVal:
```
Python 3.5.3 (default, Jan 19 2017, 14:11:04) 
[GCC 6.3.0 20170118] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import subprocess
>>> subprocess.Popen(['echo', 5])
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/subprocess.py", line 676, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1221, in _execute_child
    restore_signals, start_new_session, preexec_fn)
TypeError: Can't convert 'int' object to str implicitly
```

plugins/Unix/config.py
```diff
@@ -108,6 +108,8 @@ class SpaceSeparatedListOfNonOptionStrings(registry.SpaceSeparatedListOfStrings)
 conf.registerGlobalValue(Unix.ping, 'command', 
     registry.String(utils.findBinaryInPath('ping') or '', """Determines what 
     command will be called for the ping command."""))
+conf.registerGlobalValue(Unix.ping, 'defaultPingCount',
+    registry.Integer(5, """Determines what ping and ping6 counts(-c) will default to."""))
```
GLolol: Maybe use `registry.PositiveInteger` so values like -1 and 0 can't be used?

GLolol:
Thanks, merged with one more fix in 82cdf1796433359eff323801c7995e8f07778677

PR #1308, opened 2017-10-25
Config: Don't use a prefix in _setValue (closes #1307)

### LINKED ISSUE #1307: config supybot.commands.allowShell true suggests using commands with @ prefix
> `Error: This configuration variable is not writeable via IRC. To change it you have to: 1) use @flush 2) edit the config file 3) use @config reload.`

As far as I can see, naming `@` prefix is inconsistent and isn't used anywhere else in Limnoria. For example:

> `(scheduler add <seconds> <command>) -- Schedules the command string <command> to run <seconds> seconds in the future. For example, 'scheduler add [seconds 30m] "echo [cpu]"' will schedule the command "cpu" to be sent to the channel the schedule add command was given in (with no prefixed nick, a consequence of using echo). Do pay attention to the quotes in that example.`

* https://github.com/ProgVal/Limnoria/blob/master-2017-10-25/plugins/Scheduler/plugin.py#L136-L140

IotaSpencer:
-- deviated from normal '' line enclosure to allow use of ' in lines without escaping

### DIFF
--- a/plugins/Config/plugin.py
+++ b/plugins/Config/plugin.py
@@ -213,9 +213,9 @@ class Config(callbacks.Plugin):
 
     def _setValue(self, irc, msg, group, value):
         if isReadOnly(group._name):
-            irc.error(_('This configuration variable is not writeable '
-                'via IRC. To change it you have to: 1) use @flush 2) edit '
-                'the config file 3) use @config reload.'), Raise=True)
+            irc.error(_("This configuration variable is not writeable "
+                "via IRC. To change it you have to: 1) use 'flush' 2) edit "
+                "the config file 3) use 'config reload'."), Raise=True)
         capability = getCapability(group._name)
         if ircdb.checkCapability(msg.prefix, capability):
             # I think callCommand catches exceptions here.  Should it?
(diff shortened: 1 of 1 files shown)


plugins/Config/plugin.py
```diff
@@ -213,9 +213,9 @@ def _getValue(self, irc, msg, group, addChannel=False):
 
     def _setValue(self, irc, msg, group, value):
         if isReadOnly(group._name):
-            irc.error(_('This configuration variable is not writeable '
-                'via IRC. To change it you have to: 1) use @flush 2) edit '
-                'the config file 3) use @config reload.'), Raise=True)
+            irc.error(_("This configuration variable is not writeable "
+                "via IRC. To change it you have to: 1) use 'flush' 2) edit "
+                "the config file 3) use 'config reload'."), Raise=True)
```
Mikaela: use the 'config reload' command

Mikaela: And I would 1) flush 2) config flush false 3)edit the config file 4)config reload.

I am on mobile web and I think I failed with commenting at least partially and I don't know what is the proper way to document this or would this be too detailed.

I also think that people aren't supposed to edit the config so I feel unsure about telling them to do that.

Mikaela:
I am now happy with the changes assuming that the user is wanted to be told to edit the config file unlike most of time.

I would also add "Closes #1307" to commit message or PR text so it would automatically close the issue upon reaching the master branch.

PR #1309, opened 2017-10-25
Use a prefix-less help string, don't assume a '@' prefix

IotaSpencer:


plugins/Nickometer/locales/fi.po
```diff
@@ -18,10 +18,10 @@ msgstr ""
 "X-Generator: Poedit 1.6.10\n"
 
 #: plugin.py:81
-msgid "Will tell you how lame a nick is by the command `@nickometer [nick]`."
+msgid "Will tell you how lame a nick is by the command 'nickometer [nick]'."
```
ProgVal: you changed backticks into apostrophes in this msgid but not in the others, and you did not change the plugin.py at all.

PR #1310, opened 2017-10-26
Time: add 'ddate'

IotaSpencer:
…in setup.py


plugins/Time/local/ddate.py
```diff
@@ -0,0 +1,175 @@
+# copied from https://github.com/a-tal/ddate
```
GLolol: Is there a reason why you embed ddate instead of using it as an optional dependency? It is available on [PyPI](https://pypi.python.org/pypi/ddate/) after all.

IotaSpencer: to allow use from install, without anything else being installed

ProgVal: and this notice does not follow ddate's license:

```
Redistributions of source code must retain the above copyright notice, this
list of conditions and the following disclaimer.
```

IotaSpencer: then I'll copy that over

ProgVal: I would still prefer an optional dependency

ProgVal: I would still prefer an optional dependency, unless you see an issue with that.

ProgVal:
Could you add test cases?

plugins/Time/local/ddate.py
GLolol: Seriously though, I don't agree with embedding dependencies simply for the sake of doing so. In the long run it's more maintenance effort and makes makes tracking fixes, etc. harder. Distros are very much against this practice too for good reason:

- https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies
- https://wiki.debian.org/EmbeddedCodeCopies

plugins/Time/plugin.py
```diff
@@ -206,7 +206,17 @@ def tztime(self, irc, msg, args, timezone):
...
 
-
+    def ddate(self, irc, msg, args, year=None, month=None, day=None):
+        """[<year> <month> <day>]
+        Returns a the Discordian date today, or an optional different date."""
+        if year is not None and month is not None and day is not None:
+            try:
+                irc.reply(_ddate(datetime(year=year, month=month, day=day)))
+            except ValueError as e:
+                irc.error("{}".format(e))
```
ProgVal: `irc.error("%s", e)`

plugins/Time/local/ddate.py
IotaSpencer: I'll change it to an optional dependency

ProgVal: @GLolol None of these arguments apply to such a simple library, though.

plugins/Time/plugin.py
IotaSpencer: oh right, to maintain backwards compatibility

setup.py
```diff
@@ -152,6 +152,7 @@ def normalizeWhitespace(s):
             [
              'supybot.plugins.Dict.local',
              'supybot.plugins.Math.local',
+             'supybot.plugins.Time.local',
```
ProgVal: you can remove this

setup.py
```diff
@@ -161,6 +162,7 @@ def normalizeWhitespace(s):
                'supybot.drivers': 'src/drivers',
                'supybot.plugins.Dict.local': 'plugins/Dict/local',
                'supybot.plugins.Math.local': 'plugins/Math/local',
+               'supybot.plugins.Time.local': 'plugins/Time/local',
```
ProgVal: same

plugins/Time/plugin.py
```diff
@@ -206,7 +212,20 @@ def tztime(self, irc, msg, args, timezone):
...
+        if _ddate is not None:
+            if year is not None and month is not None and day is not None:
+                try:
+                    irc.reply(_ddate(datetime(year=year, month=month, day=day)))
+                except ValueError as e:
+                    irc.error("%s", e)
+            else:
+                irc.reply(_ddate())
+        else:
+            irc.error("The 'ddate' module is not installed. Use 'pip install --user ddate' See <https://pypi.python.org/pypi/ddate/> for more information.")
```
ProgVal: Could you split lines, wrap the string with `_()`, and use `"%s -m pip", sys.executable` instead of just `pip`?

PR #1313, opened 2017-10-28
src/ircmsgs: modes doesn't just quiet ('Returns a MODE to /quiet/ eac…

IotaSpencer:
…h of nicks on channel')

* plus also fix some possible english issues

### DIFF
--- a/src/ircmsgs.py
+++ b/src/ircmsgs.py
@@ -840,7 +840,7 @@ def mode(channel, args=(), prefix='', msg=None):
     return IrcMsg(prefix=prefix, command='MODE', args=(channel,)+args, msg=msg)
 
 def modes(channel, args=(), prefix='', msg=None):
-    """Returns a MODE to each of nicks on channel."""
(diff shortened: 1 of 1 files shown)


GLolol:
I actually find the new description just as confusing as the original... Maybe something like "Returns a MODE message for the channel for all the (mode, targetOrNone) pairs in 'args'."

src/ircmsgs.py
```diff
@@ -840,7 +840,7 @@ def mode(channel, args=(), prefix='', msg=None):
     return IrcMsg(prefix=prefix, command='MODE', args=(channel,)+args, msg=msg)
 
 def modes(channel, args=(), prefix='', msg=None):
-    """Returns a MODE to each of nicks on channel."""
+    """Returns a MODE message for the channel for all the (mode, targetOrNone) pairs in 'args'."""
```
ProgVal: these are 2-tuples, not pairs

PR #1314, opened 2017-10-29
Add a couple more crypt options

IotaSpencer:


src/utils/crypt.py
```diff
@@ -29,5 +29,8 @@
 
 from hashlib import md5
 from hashlib import sha1 as sha
+from hashlib import sha256 as sha_256
+from hashlib import sha512 as sha_512
+from hashlib import ripemd160 as ripemd_160
```
GLolol: Why not leave these names as is without the underscores?

plugins/String/plugin.py
```diff
@@ -250,6 +250,33 @@ def sha(self, irc, msg, args, text):
...
+
+    @internationalizeDocstring
+    def ripemd160(self, irc, msg, args, text):
+        """<text>
+
+        Returns a RIPEMD160 hash of the given string.
+        """
+        irc.reply(utils.crypt.ripemd_160(text.encode('utf8')).hexdigest())
+    ripemd160 = wrap(ripemd160, ['text'])
+
```
GLolol: Can you add tests?

src/utils/crypt.py
```diff
@@ -29,5 +30,8 @@
...
```
ProgVal: `ripemd160` is not guaranteed to be available on all Python systems, see [hashlib's documentated](https://docs.python.org/2/library/hashlib.html)

src/utils/crypt.py
```diff
@@ -1,5 +1,6 @@
 ###
 # Copyright (c) 2008, James McCoy
+# Copyright (c) 2017, Ken Spencer
```
ProgVal: this is a trivial change, copyright does not need to be updated

src/utils/crypt.py
IotaSpencer: true

plugins/String/plugin.py
```diff
@@ -250,6 +251,33 @@ def sha(self, irc, msg, args, text):
...
```
ProgVal: You could make a function to generate commands instead of duplicating the same code.

plugins/String/test.py
```diff
@@ -145,6 +145,12 @@ def testXor(self):
         for s0, s1, s2, s3, s4, s5, s6, s7, s8, s9 in group(L, 10):
             data = '%s%s%s%s%s%s%s%s%s' % (s0, s1, s2, s3, s4, s5, s6, s7, s8)
             self.assertResponse('xor %s [xor %s %s]' % (s9, s9, data), data)
+    brown_fox_string = 'The quick brown fox jumped over the lazy dog'
```
ProgVal: unused string

plugins/String/plugin.py
```diff
@@ -240,15 +241,42 @@ def md5(self, irc, msg, args, text):
...
+        irc.reply(utils.crypt.sha_512(text.encode('utf8')).hexdigest())
+    sha512 = wrap(sha512, ['text'])
+
+    @internationalizeDocstring
+    def ripemd160(self, irc, msg, args, text):
+        """<text>
+
+        Returns a RIPEMD160 hash of the given string.
+        """
+        irc.reply(utils.crypt.new('ripemd160', text.encode('utf8')).hexdigest())
```
ProgVal: ripemd160 is not guaranteed to be available

src/utils/__init__.py
```diff
@@ -62,6 +62,6 @@ def force(x):
 # These imports need to happen below the block above, so things get put into
 # __builtins__ appropriately.
 from .gen import *
-from . import crypt, error, file, iter, net, python, seq, str, transaction, web
+from . import error, file, iter, net, python, seq, str, transaction, web
```
ProgVal: no, that's a breaking change

IotaSpencer:
The scale of this PR is getting ridiculous, closing

PR #1315, opened 2017-10-29
Add hashes plugin

IotaSpencer:


plugins/Hashes/plugin.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        irc.reply(hashlib.sha256(text.encode('utf8')).hexdigest())
+    sha512 = wrap(sha512, ['text'])
+
+    @internationalizeDocstring
+    def algorithms(self, irc, msg, args):
+        """<takes no arguments>
+
+        Returns the list of available algorithms."""
+        algos = list(hashlib.algorithms_available)
```
GLolol: `irc.reply(format("%L", hashlib.algorithms_available))`

plugins/Hashes/plugin.py
```diff
@@ -0,0 +1,129 @@
...
+        irc.reply("%s" % algos.join(', '))
+    algorithms = wrap(algorithms)
+
+    @internationalizeDocstring
+    def mkhash(self, irc, msg, args, algorithm, text):
+        """<algorithm> <text>
+
+        Returns TEXT after it has been hashed with ALGORITHM. See the 'algorithms' command in this plugin to return the algorithms available on this system."""
+        algos = list(hashlib.algorithms_available)
```
GLolol: You shouldn't need the extra `list()`

plugins/Hashes/plugin.py
```diff
@@ -111,16 +111,15 @@ def algorithms(self, irc, msg, args):
         """<takes no arguments>
 
         Returns the list of available algorithms."""
-        algos = list(hashlib.algorithms_available)
-        irc.reply("%s" % algos.join(', '))
+        irc.reply("%L" % hashlib.algorithms_available)
```
GLolol: No, `format("%L", hashlib.algorithms_available)`. You need to use Limnoria's format function from supybot.utils.str

GLolol:
Two more things:
- Please fix the tests!
- I don't understand why the tests for xor are so complicated. Can't we just test them with simple strings?

plugins/Hashes/plugin.py
```diff
@@ -0,0 +1,117 @@
...
+    @internationalizeDocstring
+    def mkhash(self, irc, msg, args, algorithm, text):
+        """<algorithm> <text>
+
+        Returns TEXT after it has been hashed with ALGORITHM. See the 'algorithms' command in this plugin to return the algorithms available on this system."""
+        algos = hashlib.algorithms_available
+        if algorithm not in algos:
+            irc.error("Algorithm not available.")
+        else:
+            irc.reply(hashlib.new("%s" % algorithm, b"%s" % text.encode('utf8')).hexdigest())
```
GLolol: No need for string substitution at all if all you're adding in is a variable

plugins/Hashes/plugin.py
```diff
@@ -112,6 +112,6 @@ def mkhash(self, irc, msg, args, algorithm, text):
         if algorithm not in algos:
             irc.error("Algorithm not available.")
         else:
-            irc.reply(hashlib.new("%s" % algorithm, b"%s" % text.encode('utf8')).hexdigest())
+            irc.reply(hashlib.new(algorithm, b"%s" % text.encode('utf8')).hexdigest())
```
GLolol: Same with the b"%s" part

plugins/Hashes/plugin.py
```diff
@@ -0,0 +1,127 @@
...
+
+def registerDefaultPlugin(command, plugin):
+    command = callbacks.canonicalName(command)
+    conf.registerGlobalValue(conf.supybot.commands.defaultPlugins,
+                             command, registry.String(plugin, ''))
+    # This must be set, or the quotes won't be removed.
+    conf.supybot.commands.defaultPlugins.get(command).set(plugin)
+
+registerDefaultPlugin('md5', 'Hashes')
+registerDefaultPlugin('sha', 'Hashes')
```
ProgVal: no, registerDefaultPlugin should only be used for important/core commands.

PR #1317, opened 2017-12-10
Socket: exempt local IPs from "not using TLS" warnings

GLolol:
This is very minor, but I run a bot that connects to localhost where enabling TLS isn't very important. That said, I'm not sure if this is the most versatile check or whether this is a good place to put it (versus abstracting it away somewhere).


ProgVal:
```
try:
    import ipaddress
except ImportError:
    # Python < 3.3
    ipaddress = None

# ...

if ipaddress is None or not ipaddress.Address(address).is_loopback:
```

PR #1327, opened 2018-01-16
Use of the most usefull arguments for WHOX

ncoevoet:
in order to avoid external plugins to do again the same request with more arguments, as thiis request must be made per channel and can take a lot of time, it's rather better if core limnoria already shares all needed informations

### DIFF
--- a/plugins/NickAuth/plugin.py
+++ b/plugins/NickAuth/plugin.py
@@ -199,10 +199,11 @@ class NickAuth(callbacks.Plugin):
             self._auth(irc, msg.prefix, account)
 
     def do354(self, irc, msg):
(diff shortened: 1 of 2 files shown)


ProgVal:
You should change NickAuth too, see https://github.com/ProgVal/Limnoria/commit/20c71df8f9689da4e0bb5001644bb724f6a48776

ProgVal:
Also, it looks like you put `ip` in the wrong place when unpacking

ncoevoet:
the ip is on the right place, i just tested it again, i will do another pull request for NickAuth

ProgVal:
Oh, indeed, I misunderstood the format.

Do all servers with WHOX support all of these characters?
If not, that would raise an exception when unpacking.

PR #1334, opened 2018-04-04
setup.py: Remove unneeded import of match for ceil() function

JohnVillalovos:
The ceil() function was previously used in setup.py was removed by
commit c3a2c800f1c36adff0db9651c2d74c3f6a648f84. So we no longer need
it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,6 @@ import warnings
 import datetime
 import tempfile
 import subprocess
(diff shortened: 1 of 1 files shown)


GLolol:
Pull requests should be sent to testing so I've gone ahead and switched the base branch. Other than that, looks good - thanks for your contribution!

PR #1338, opened 2018-06-01
Allow to override build date

bmwiedemann:
Allow to override build date
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,8 +62,9 @@ except:
     if os.path.isfile(VERSION_FILE):
         from src.version import version
     else:
-        from time import gmtime, strftime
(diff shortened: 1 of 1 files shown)


ProgVal:
Hmm, if `SOURCE_DATE_EPOCH` is available, it should be used in the first branch of the `try`. I'll fix that.

bmwiedemann:
I think it was OK to use (reproducible) git commit time. just using the 'current' build time is problematic.

ProgVal:
Using the current build time is a fallback in case the Git history is not available.

PR #1340, opened 2018-06-06
Fixes and tests to supybot-plugin-create; modernize the plugin template

jlu5:
- cb33687: compactify import statements where applicable into `from supybot import abc, def, ...`
- af515ec: prefer `importlib` over `imp` on Python >= 3.4, where `imp` is deprecated

### DIFF
--- a/scripts/supybot-plugin-create
+++ b/scripts/supybot-plugin-create
@@ -153,6 +153,7 @@ __init__Template = '''
 %s: %s
 """
 
+import sys
 import supybot
 from supybot import world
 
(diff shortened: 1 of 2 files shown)


ProgVal:
Did you check it works for all python versions?

jlu5:
That's the only thing I'm not sure about, and it'd be a tedious process to manually test. Some thoughts though:

1) It looks like multiline wrapping for multiple imports (`from a import b, c, ...`) was [refined in Python 2.4-2.6](https://www.python.org/dev/peps/pep-0328/), leading me to assume that this is supported by even earlier versions?

2) I'm also assuming that new plugins created going forward will likely not be targeting super old Python versions, given the fact that Python 2 is EOL in a couple of years.

IotaSpencer:
I would suggest also putting in a commented out method so you can figure out the syntax, including making sure there is `def function(self, irc, msg, args, OTHER, ARGS):` as well as an example docstring since those are required.

ProgVal:
LGTM. Can I squash&merge?

PR #1343, opened 2018-08-03
Fix time namespaces in setup.py when version file is unavailable

mrdaemon:
Running setup.py in environments where:

1) SOURCE_DATE_EPOCH is unavailable
2) git is not available
3) The version file is not present

Will die with:
```
  File "setup.py", line 69, in <module>
    version = 'installed on ' + strftime("%Y-%m-%dT%H-%M-%S", gmtime())
NameError: name 'strftime' is not defined
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ except:
     if os.path.isfile(VERSION_FILE):
         from src.version import version
     else:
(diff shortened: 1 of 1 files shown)


ProgVal:
For some reason I removed the import in 36309ad1dcaf6e86befab48d7226ce3ad4d00365...

Thanks for the patch!

PR #1344, opened 2018-09-01
Small fix on french translation for kban

architek:
One word missing in french translation for kban


jlu5:
Thanks for your contribution!

PR #1345, opened 2018-09-07
Explicitly convert ipaddress to unicode for ipaddress Python 2 backwards compatibility

averi:
This fixes:

  File "/usr/lib/python2.7/site-packages/supybot/drivers/Socket.py", line 289, in reconnect
    (ipaddress is None or not ipaddress.ip_address(address).is_loopback):
  File "/usr/lib/python2.7/site-packages/ipaddress.py", line 163, in ip_address
    ' a unicode object?' % address)
AddressValueError: 'x.x.x.x' does not appear to be an IPv4 or IPv6 address. Did you pass in a bytes (str in Python 2) instead of a unicode object?

### DIFF
--- a/src/drivers/Socket.py
+++ b/src/drivers/Socket.py
@@ -286,7 +286,7 @@ class SocketDriver(drivers.IrcDriver, drivers.ServersMixin):
                 self.starttls()
             # Suppress this warning for loopback IPs.
             elif (not network_config.requireStarttls()) and \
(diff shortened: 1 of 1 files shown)


ProgVal:
Hi and thanks for the PR

I believe that bug was fixed in c5175110e01033c99c18bbbcf9cc4d7abe0d67e0 (which is not released yet).

PR #1370, opened 2019-09-17
only raise in command.py.process() when a raise was wrapped

jesopo:
In the unlikely event that a command `return`s an Exception (instead of `raise`ing it), it should not be raised up out of command.py.process()


ProgVal:
Nitpick: I'd rather you flipped the boolean, and named the variable `raised` instead, I think it's clearer than what a success it

PR #1371, opened 2019-09-30
Fix 'Uncaught exception', makes check for 'charade' dep to logs, others

### LINKED ISSUE #1359: Page (possibly UTF content) makes plugins/web/titleSnarfer breaks (Uncaught exception)
Hi. While trying to check why titleSnarfer won't return Title's page for http://lastsummer.de/creating-custom-packages-on-freebsd found this on the logs:

```
ERROR 2019-01-31T18:23:44 supybot Printed to stderr after daemonization: Exception in thread Thread #62 (for snarfing http://lastsummer.de/creating-custom-packages-on-freebsd/):
   File "/usr/local/lib/python2.7/threading.py", line 801, in __bootstrap_inner
     self.run()
   File "/usr/local/lib/python2.7/site-packages/supybot/commands.py", line 166, in run
     super(UrlSnarfThread, self).run()
   File "/usr/local/lib/python2.7/threading.py", line 754, in run
     self.__target(*self.__args, **self.__kwargs)
   File "/usr/local/lib/python2.7/site-packages/supybot/commands.py", line 217, in doSnarf
     f(self, irc, msg, match, *L, **kwargs)
   File "/usr/local/lib/python2.7/site-packages/supybot/plugins/Web/plugin.py", line 194, in titleSnarfer
     r = self.getTitle(irc, url, False)
   File "/usr/local/lib/python2.7/site-packages/supybot/plugins/Web/plugin.py", line 166, in getTitle
     parser.feed(text)
   File "/usr/local/lib/python2.7/HTMLParser.py", line 117, in feed
     self.goahead(0)
   File "/usr/local/lib/python2.7/HTMLParser.py", line 161, in goahead
     k = self.parse_starttag(i)
   File "/usr/local/lib/python2.7/HTMLParser.py", line 308, in parse_starttag
     attrvalue = self.unescape(attrvalue)
   File "/usr/local/lib/python2.7/HTMLParser.py", line 476, in unescape
     return re.sub(r"&(#?[xX]?(?:[0-9a-fA-F]+|\w{1,8}));", replaceEntities, s)
   File "/usr/local/lib/python2.7/re.py", line 155, in sub
     return _compile(pattern, flags).sub(repl, string, count)
 UnicodeDecodeError: 'ascii' codec can't decode byte 0xbb in position 0: ordinal not in range(128)
 INFO 2019-01-31T18:23:45 supybot Flushers flushed and garbage collected.
 ERROR 2019-01-31T18:24:15 supybot Unhandled error message from server: IrcMsg(prefix="cherryh.freenode.net", command="412", args=('VimDiesel', 'No text to send'))
 ERROR 2019-01-31T18:24:16 supybot Unhandled error message from server: IrcMsg(prefix="cherryh.freenode.net", command="412", args=('VimDiesel', 'No text to send'))
```

In this block from plugins/Web/plugin.py:

```
try:
            text = text.decode(utils.web.getEncoding(text) or 'utf8',
                    'replace')
        except UnicodeDecodeError:
            pass
        parser = Title()
        if minisix.PY3 and isinstance(text, bytes):
            if raiseErrors:
                irc.error(_('Could not guess the page\'s encoding. (Try '
                        'installing python-charade.)'), Raise=True)
            else:
                return None
        parser.feed(text)
        parser.close()
```

It seems that 'text' is the whole page to be parsed by HTMLparser. Anyway, changing line 166 (in my plugin.py copy) from
parser.feed(text)
to
parser.feed(text.encode('utf-8'))

Fixed the problem for this specific page while other pages (as far as I tested) keeps working as usual.
Can't conclude what is the problem or how relevant it could be, reporting here in case this example is useful.

The current (running) version of this Limnoria is installed on 2018-12-22T04-00-46, running on Python 2.7.15 (default, Dec 20 2018, 01:13:53)  [GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)].  The newest versions available online are 2019.01.27 (in testing), 2018.12.19 (in master).
Salute!

### LINKED ISSUE #1362: utils.web.getEncoding() always returning 'None' in Web plugin
Hi. While trying to find why NBSP (non-breaking space) decodes incorrectly if page is charset iso8859-1 I discovered that in the Web plugin, actual line 155 "text = text.decode(utils.web.getEncoding(text) or 'utf8',            'replace')" the utils.web.getEncoding(text) is always returning 'None'.
I tried a couple of different pages with same result, getEnconding not being capable of returning actual encoding.
Example of the problem: Title returned in the page 'https://www.freebsd.org/doc/handbook/usb-device-mode-terminals.html' the title contains nbsp in the right encoding accordingly iso8859-1.

I didn't look at getEnconding() yet to try finding the issue (in the case it's really a getEnconding())

The current (running) version of this Limnoria is installed on 2019-01-24T22-12-03, running on Python 3.6.8 (default, Jan  3 2019, 01:10:23)  [GCC 4.2.1 Compatible FreeBSD Clang 6.0.0 (tags/RELEASE_600/final 326565)].  The newest versions available online are 2019.02.22 (in master), 2019.02.22 (in testing).

Thanks!

Rodrigo-NH:
Despite closed, uncaught exceptions still present for both py2 and py3 on generic exceptions (ssl mismatches, 404, 403 etc) not caught by utils/web 'getUrlFd' and breaking 'getTitle'. Unbreak and make those exceptions to log instead, informing the offending URL. Testing, I saw some exceptions will cause 'getTitle' threads to block and never get past exception handles. Including explicit evaluation and error handling of '(target, text)' inside owns 'getTitle' fixed this blocking on my tests.
Fixes #1359

utils.web.getEncoding() always returning 'None' in Web plugin
Seems that the way it's now, message 'Try installing python-charade' never get to user knowledge trough logs (perhaps cause of issues described #1359), and try informing only in case of 'minisix.PY3', but also important for Py2. Makes explicit test for charade and inform to logs.
Fixes #1362

Testing: temporarily wrapped the code in a 'while' that loaded and processed 470 mixed urls in both py2 and py3, worked as expected. Tested parsing links in IRC and watching, produced same results for both py2 and py3.

### DIFF
--- a/src/ircmsgs.py
+++ b/src/ircmsgs.py
@@ -73,7 +73,12 @@ def parse_server_tags(s):
             server_tags[tag] = None
         else:
             (key, value) = tag.split('=', 1)
-            server_tags[key] = unescape_server_tag_value(value)
+            value = unescape_server_tag_value(value)
+            if value == '':
+                # "Implementations MUST interpret empty tag values (e.g. foo=)
+                # as equivalent to missing tag values (e.g. foo)."
+                value = None
+            server_tags[key] = value
     return server_tags
 def format_server_tags(server_tags):
     parts = []
--- a/test/test_ircmsgs.py
+++ b/test/test_ircmsgs.py
@@ -147,6 +147,14 @@ class IrcMsgTestCase(SupyTestCase):
         self.assertEqual(m.args, ('me', 'Hello'))
         self.assertEqual(str(m), s + '\n')
 
+        s = '@foo=;bar=baz;qux= ' \
+            ':nick!ident@host.com PRIVMSG me :Hello'
+        m = ircmsgs.IrcMsg(s)
+        self.assertEqual(m.server_tags, {
+            'foo': None,
+            'bar': 'baz',
+            'qux': None})
+
     def testTime(self):
         before = time.time()
         msg = ircmsgs.IrcMsg('PRIVMSG #foo :foo')


plugins/Web/plugin.py
```diff
@@ -149,38 +149,31 @@ def noIgnore(self, irc, msg):
...
-        if minisix.PY3 and isinstance(text, bytes):
-            if raiseErrors:
-                irc.error(_('Could not guess the page\'s encoding. (Try '
-                        'installing python-charade.)'), Raise=True)
-            else:
-                return None
         try:
+            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,timeout=timeout)
+            encoding = utils.web.getEncoding(text)
+            if encoding is None: # Condition if charade not installed
```
ProgVal: `utils.web.getEncoding` may return None even if charade is not installed. The log line should be in `utils.web.getEncoding` instead.

plugins/Web/plugin.py
```diff
@@ -149,38 +149,31 @@ def noIgnore(self, irc, msg):
...
+                self.log.info('Web plugin TitleSnarfer: Could not guess the page\'s'
+                              ' encoding. (Try installing python-charade.)')
+                encoding = 'utf-8' # Assume UTF-8 and replace unknown chars to the UTF-8 codec for U+FFFD in the next hop
+            text = text.decode(utils.web.getEncoding(text) or 'utf-8','replace')
```
ProgVal: replace `utils.web.getEncoding(text) or 'utf-8'` with `encoding`, or remove the previous lines.

plugins/Web/plugin.py
```diff
@@ -149,38 +149,31 @@ def noIgnore(self, irc, msg):
...
-            return (target, title)
-        elif raiseErrors:
-            if len(text) < size:
-                irc.error(_('That URL appears to have no HTML title.'),
-                        Raise=True)
+            try:
+                parser.feed(text)
+            except:
+                parser = Title()
+                parser.feed(bytes(text)) # Explicitly pack to bytes in encoding errors for (more) python2 compatibility
```
ProgVal: As it's only for py2, I'd like this to be guarded by a Python version check.

plugins/Web/plugin.py
```diff
@@ -149,38 +149,31 @@ def noIgnore(self, irc, msg):
...
+                return (target, title)
             else:
-                irc.error(format(_('That URL appears to have no HTML title '
-                                 'within the first %S.'), size), Raise=True)
+                if len(text) < size:
+                    self.log.info('Web plugin TitleSnarfer: <' + url + '> appears to have no HTML title.')
+                else:
+                    self.log.info('Web plugin TitleSnarfer: Could not retrieve title of <' + url + '>')
+        except Exception as e:
+            self.log.info('Web plugin TitleSnarfer: <' + str(e) + '> while trying to process <' + url +'>')
```
ProgVal: This should raise an error if `raiseErrors` is True (like it did before).

Also, the proper way to use string substitution for logs is:

```
self.log.info('Web plugin TitleSnarfer: <%s> appears to have no HTML title.', url)
```

ProgVal:
hmm?

PR #1373, opened 2019-10-05
Aka: Add web UI

kyrias:


src/httpserver.py
```diff
@@ -45,8 +45,10 @@
 
 if minisix.PY2:
     from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+    from cgi import escape
 else:
     from http.server import HTTPServer, BaseHTTPRequestHandler
+    from html import escape
```
ProgVal: Should probably be renamed `html_escape` and moved to `utils/web.py`.

plugins/Aka/plugin.py
```diff
@@ -376,6 +396,103 @@ def findBiggestAt(alias):
...
+                        <td style="white-space: nowrap;"><code>{name}</code></td>
+                        <td><code>{alias}<code></td>
+                        <td style="white-space: nowrap;">{locked}</td>
+                    </tr>
+                """.format(
+                    name=httpserver.escape(name),
+                    alias=httpserver.escape(alias),
+                    locked=httpserver.escape(locked_column)),
+                )
+            self.write(self.list_template.format('\n'.join(aka_rows)))
```
ProgVal: doesn't need a `\n`

plugins/Aka/plugin.py
```diff
@@ -376,6 +396,104 @@ def findBiggestAt(alias):
...
+                    akas[aka[1]] = aka
+
+            aka_rows = []
+            for name in sorted(akas.keys()):
+                (id, name, alias, locked, locked_by, locked_at) = akas[name]
+                locked_column = 'False'
+                if locked:
+                    locked_column = 'By {} at {}'.format(
+                        locked_by,
+                        locked_at.split('.')[0],
```
ProgVal: `utils.str.timestamp(float(locked_at))`

plugins/Aka/plugin.py
```diff
@@ -376,6 +396,104 @@ def findBiggestAt(alias):
...
+            for aka in self._plugin._db.get_all('global'):
+                akas[aka[1]] = aka
+
+            if channel != 'global':
+                for aka in self._plugin._db.get_all(channel):
+                    akas[aka[1]] = aka
+
+            aka_rows = []
+            for name in sorted(akas.keys()):
+                (id, name, alias, locked, locked_by, locked_at) = akas[name]
```
ProgVal:
```
for (name, aka) in sorted(akas.items()):
    (id, name, alias, locked, locked_by, locked_at) = aka
```

plugins/Aka/plugin.py
kyrias: They're stored as strings like `2019-10-05 22:40:03.531491` in SQLite3 though.

kyrias:
I'll try to write up some tests for it tomorrow!

plugins/Aka/plugin.py
```diff
@@ -193,6 +194,15 @@ def get_aka_lock(self, channel, name):
...
+        def get_all(self, channel):
+            cursor = self.get_db(channel).cursor()
+            cursor.execute("""
+                SELECT id, name, alias, locked, locked_by, locked_at
+                FROM aliases;
+            """)
+            return cursor.fetchall()
+
     available_db.update({'sqlite3': SQLiteAkaDB})
 elif sqlalchemy:
```
kyrias: Additionally, this is dead code, since sqlite3 has been a core module since Python 2.5.

ProgVal: Isn't it optionally part of python, though? Anyway, I think you're right, I want to get rid of the sqlalchemy backend too; it's just duplicate code with barely any benefit.

plugins/Aka/plugin.py
```diff
@@ -376,6 +396,104 @@ def findBiggestAt(alias):
...
+<html>
+  <head>
+    <title>Aka</title>
+    <link rel="stylesheet" href="/default.css">
+  </head>
+
+  <body>
+    <h1>Aka</h1>
+
+    {}
```
kyrias: So Python 2.6 is complaining that this doesn't have an explicit `{0}`, but since it went EOL in 2013, do we need to care?

ProgVal: I'll remove its support around 2020-01-01, at the same time as 2.7 and probably 3.2 and 3.3 as well. In the meantime, could you use `%s`, for consistency with other parts of Limnoria?

plugins/Aka/plugin.py
kyrias: Oh, indeed it is, though I've never seen any build without it, and everything I've seen in the last ~5 years always expect it to be there.

plugins/Aka/plugin.py
kyrias: Ah, okay, will do.

plugins/Aka/plugin.py
ProgVal: oh, ok

plugins/Aka/test.py
```diff
@@ -292,4 +293,48 @@ def testSearch(self):
...
+    def testChannelPage(self):
+        self.assertNotError('config plugins.Aka.web.enable True')
+
+        self.assertNotError('aka add foo1 echo 1')
+        self.assertNotError('aka add --channel #foo foo2 echo 2')
+
+        (respCode, body) = self.request('/aka/list/%23foo')
+        self.assertEqual(respCode, 200)
+        self.assertIn(b'foo1', body)
+        self.assertIn(b'foo2', body)
```
ProgVal: also check that an aka displayed for another channel is not shown on that page

plugins/Aka/plugin.py
kyrias: It is probably worth looking into passing `sqlite3.PARSE_DECLTYPES | sqlite3.PARSE_COLNAMES` as the `detect_types` argument to `sqlite3.connect` option to get back an actual `datetime` object, but that should really be done as a separate thing.

src/test.py
```diff
@@ -653,14 +651,14 @@ def setUp(self):
...
-        rfile = minisix.io.StringIO()
+        wfile = minisix.io.BytesIO()
+        rfile = minisix.io.BytesIO()
         connection = FakeHTTPConnection(wfile, rfile)
         connection.putrequest(method, url)
         connection.endheaders()
         rfile.seek(0)
-        wfile.seek(0)
         handler = TestRequestHandler(rfile, wfile)
+        wfile.seek(0)
```
ProgVal: wow how did that even work before that commit

kyrias: I don't think it ever did, these classes just aren't used.  All the code that would be required to write testcases with them is there, but as far as I can tell, they never ended up being used in actual test cases.

ProgVal: They are used in the WebStats plugin in https://github.com/ProgVal/Supybot-plugins/ (although its tests are disabled on py3 because I was too lazy to look into them)

ProgVal: hmm but you're right, those tests are really basic, so they just happened to work.

PR #1376, opened 2019-10-06
RSS: add 'announce chans' command to list channels a feed is announced to

### LINKED ISSUE #1322: Add option to see which channels an RSS feed announces to
Currently (and I've searched through `config search` and the `@rss command itself`) we don't have a way to know which channels are receiving announcements of a specific RSS. would be nice something like `@rss announce chanlist <feed_name>` to see for which channels that feed is being sent (a easier way to then use `@rss announce remove #channel <feed_name>`).

Best regards

jlu5:
I'm assuming `get_feed()` is the right call here. This seems to work fine with both name and URL added feeds on my bot.

Closes #1322.

### DIFF
--- a/plugins/RSS/plugin.py
+++ b/plugins/RSS/plugin.py
@@ -568,9 +568,9 @@ class RSS(callbacks.Plugin):
                         channels.append(ircnet.network + channel)
 
             if channels:
-                irc.reply(format("%s announces to %L.", feed, channels))
+                irc.reply(format("%s is announced to %L.", feed, channels))
             else:
-                irc.reply("%s does not announce to any channels." % feed)
+                irc.reply("%s is not announced to any channels." % feed)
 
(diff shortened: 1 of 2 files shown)


plugins/RSS/plugin.py
```diff
@@ -550,6 +550,30 @@ def remove(self, irc, msg, args, channel, feeds):
...
+            channels = []
+            for ircnet in world.ircs:
+                for channel in ircnet.state.channels:
+                    if feed in plugin.registryValue('announce', channel, ircnet.network):
+                        channels.append(ircnet.network + channel)
+
+            if channels:
+                irc.reply(format("%s announces to %L.", feed, channels))
+            else:
+                irc.reply("%s does not announce to any channels." % feed)
```
ProgVal: nitpick: `'%s is announced to %L'` and `'%s is not announced to any channels'` for consistency

PR #1377, opened 2019-10-09
CatchTitleSnarferErrors

Rodrigo-NH:
Hi. Sorry about ditching the original pull request #1371 After doing it wrongly against 'master' and being moved to 'testing' I think that starting from an original branch seems to be the right thing. Plus after reviewing your comments there I concluded that this pull request is a better approach.
Not sure why git ended by replacing entire contiguous blocks because the changes against master aren't so big if you compare.
Description:
- Adds general error catching (404,403,ssl mismatches etc) to logs and validates 'text' for further processing.
      "  try:
            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
                timeout=timeout)
        except Exception as e:
            self.log.info('Web plugin TitleSnarfer: URL <%s> raised <%s>', url, str(e))
            text = None
        if text: "

- This is the best approach i could find with the higher success rate in case of previous encoding/decoding errors in the case of PY2
"                parser.feed(bytes(text)) # bytes() = str() in py2"

- Send information of these cases to log when no raiseErrors/title command
"           elif not raiseErrors:
                if len(text) < size:
                    self.log.info('Web plugin TitleSnarfer: URL <%s> appears'
                                  ' to have no HTML title. ', url)
                else:
                    self.log.info('Web plugin TitleSnarfer: URL <%s> appears to have no HTML title'
                                  ' within the first %S.', url, size)"
 
## Notes
- This pull request doesn't solve #1362. The issue is a dependence instead of a 'getTitle' code

Thanks!

### DIFF
--- a/plugins/Web/plugin.py
+++ b/plugins/Web/plugin.py
@@ -154,45 +154,44 @@ class Web(callbacks.PluginRegexp):
                 timeout=timeout)
         except Exception as e:
             self.log.info('Web plugin TitleSnarfer: URL <%s> raised <%s>', url, str(e))
-            text = None
-        if text:
-            try:
-                text = text.decode(utils.web.getEncoding(text) or 'utf8',
-                        'replace')
-            except UnicodeDecodeError:
-                pass
-            if minisix.PY3 and isinstance(text, bytes):
-                if raiseErrors:
-                    irc.error(_('Could not guess the page\'s encoding. (Try '
-                            'installing python-charade.)'), Raise=True)
-                else:
-                    return None
-            try:
-                parser = Title()
-                parser.feed(text)
-            except UnicodeDecodeError:
-                # Workaround for Python 2
-                # https://github.com/ProgVal/Limnoria/issues/1359
-                parser = Title()
-                parser.feed(bytes(text)) # bytes() = str() in py2
-            parser.close()
-            title = utils.str.normalizeWhitespace(''.join(parser.data).strip())
-            if title:
-                return (target, title)
-            elif raiseErrors:
-                if len(text) < size:
-                    irc.error(_('That URL appears to have no HTML title.'),
-                            Raise=True)
-                else:
-                    irc.error(format(_('That URL appears to have no HTML title '
-                                     'within the first %S.'), size), Raise=True)
-            elif not raiseErrors:
-                if len(text) < size:
-                    self.log.info('Web plugin TitleSnarfer: URL <%s> appears'
-                                  ' to have no HTML title. ', url)
-                else:
-                    self.log.info('Web plugin TitleSnarfer: URL <%s> appears to have no HTML title'
-                                  ' within the first %S.', url, size)
+            return None
+        try:
+            text = text.decode(utils.web.getEncoding(text) or 'utf8',
+                    'replace')
+        except UnicodeDecodeError:
+            pass
+        if minisix.PY3 and isinstance(text, bytes):
+            if raiseErrors:
+                irc.error(_('Could not guess the page\'s encoding. (Try '
+                        'installing python-charade.)'), Raise=True)
+            else:
+                return None
+        try:
+            parser = Title()
+            parser.feed(text)
+        except UnicodeDecodeError:
+            # Workaround for Python 2
+            # https://github.com/ProgVal/Limnoria/issues/1359
+            parser = Title()
+            parser.feed(bytes(text)) # bytes() = str() in py2
+        parser.close()
+        title = utils.str.normalizeWhitespace(''.join(parser.data).strip())
+        if title:
+            return (target, title)
+        elif raiseErrors:
+            if len(text) < size:
+                irc.error(_('That URL appears to have no HTML title.'),
+                        Raise=True)
+            else:
+                irc.error(format(_('That URL appears to have no HTML title '
+                                 'within the first %S.'), size), Raise=True)
+        elif not raiseErrors:
+            if len(text) < size:
+                self.log.info('Web plugin TitleSnarfer: URL <%s> appears'
+                              ' to have no HTML title. ', url)
+            else:
+                self.log.info('Web plugin TitleSnarfer: URL <%s> appears to have no HTML title'
+                              ' within the first %S.', url, size)
 
     @fetch_sandbox
     def titleSnarfer(self, irc, msg, match):
--- a/plugins/Web/plugin.py
+++ b/plugins/Web/plugin.py
@@ -149,38 +149,50 @@ class Web(callbacks.PluginRegexp):
     def getTitle(self, irc, url, raiseErrors):
         size = conf.supybot.protocols.http.peekSize()
         timeout = self.registryValue('timeout')
-        (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
+        try:
+            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
                 timeout=timeout)
-        try:
-            text = text.decode(utils.web.getEncoding(text) or 'utf8',
-                    'replace')
-        except UnicodeDecodeError:
-            pass
-        if minisix.PY3 and isinstance(text, bytes):
-            if raiseErrors:
-                irc.error(_('Could not guess the page\'s encoding. (Try '
-                        'installing python-charade.)'), Raise=True)
-            else:
-                return None
-        try:
-            parser = Title()
-            parser.feed(text)
-        except UnicodeDecodeError:
-            # Workaround for Python 2
-            # https://github.com/ProgVal/Limnoria/issues/1359
-            parser = Title()
-            parser.feed(text.encode('utf8'))
-        parser.close()
-        title = utils.str.normalizeWhitespace(''.join(parser.data).strip())
-        if title:
-            return (target, title)
-        elif raiseErrors:
-            if len(text) < size:
-                irc.error(_('That URL appears to have no HTML title.'),
-                        Raise=True)
-            else:
-                irc.error(format(_('That URL appears to have no HTML title '
-                                 'within the first %S.'), size), Raise=True)
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


jlu5:
(Pull requests should be sent to the testing branch)

plugins/Web/plugin.py
```diff
@@ -149,38 +149,50 @@ def noIgnore(self, irc, msg):
...
-                irc.error(_('That URL appears to have no HTML title.'),
-                        Raise=True)
-            else:
-                irc.error(format(_('That URL appears to have no HTML title '
-                                 'within the first %S.'), size), Raise=True)
+            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
+                timeout=timeout)
+        except Exception as e:
+            self.log.info('Web plugin TitleSnarfer: URL <%s> raised <%s>', url, str(e))
+            text = None
```
ProgVal: Make this a `return`, so you can drop the `if text` below

plugins/Web/plugin.py
```diff
@@ -149,38 +149,50 @@ def noIgnore(self, irc, msg):
...
+        if text:
+            try:
+                text = text.decode(utils.web.getEncoding(text) or 'utf8',
+                        'replace')
+            except UnicodeDecodeError:
+                pass
+            if minisix.PY3 and isinstance(text, bytes):
```
ProgVal: this condition should be in the `except` clause (and so you can drop the `isinstance`).

plugins/Web/plugin.py
```diff
@@ -149,38 +149,50 @@ def noIgnore(self, irc, msg):
...
-            if len(text) < size:
-                irc.error(_('That URL appears to have no HTML title.'),
-                        Raise=True)
-            else:
-                irc.error(format(_('That URL appears to have no HTML title '
-                                 'within the first %S.'), size), Raise=True)
+            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
+                timeout=timeout)
+        except Exception as e:
+            self.log.info('Web plugin TitleSnarfer: URL <%s> raised <%s>', url, str(e))
```
ProgVal: the exception should be raised if `raiseErrors`.

plugins/Web/plugin.py
```diff
@@ -149,38 +149,50 @@ def noIgnore(self, irc, msg):
...
+                else:
+                    return None
+            try:
+                parser = Title()
+                parser.feed(text)
+            except UnicodeDecodeError:
+                # Workaround for Python 2
+                # https://github.com/ProgVal/Limnoria/issues/1359
+                parser = Title()
+                parser.feed(bytes(text)) # bytes() = str() in py2
```
ProgVal: That's re-encoding a decoded string. You could use the original value of `test` instead (ie. the one before `text.decode`)

ProgVal:
Could you give me a list of URLs that get "fixed" with this PR? I'll try to write tests for those, to prevent regressions. (Even better if you write tests yourself, but I don't mind doing it :) )

plugins/Web/plugin.py
Rodrigo-NH: Ok nice! It is really much better. I will change it.

plugins/Web/plugin.py
Rodrigo-NH: Git doesn't helped in showing the changes I have made so just to make it clear: I didn't changed this part in this pull request, it's the same as before. Of course I can try changing it from now, just tell me if it's better if I change it as part of this pull request, create another after this one or if I leave it alone.

plugins/Web/plugin.py
Rodrigo-NH: Ok. Then I think it's a good idea that I add also the second case, like I did at the end of the code with 'elif not raiseErrors:' and send these to logs. This way URLs with errors can be noticed for further investigation or at least noticed with the cause titlesnarfer not returning results. Please tell me if you think its good this way.

plugins/Web/plugin.py
Rodrigo-NH: I changed this from 'parser.feed(text.encode('utf8'))' that was also re encoding things. Not sure but after reading about py2 it seems that there's no difference between this change or how it's actually, so maybe just leave it alone? I'm really unsure about this one, just know the first change fixed things for a unique URL as discussed in #1359. I can perform like you suggest but im far away from understading the logic, maybe changing like you suggest we will return to the situation before #1359? I'm really lost with this one.

Rodrigo-NH:
This is a list with some (43) URLs that get "fixed" with this PR. Actually, URLs aren't fixed in this PR but general failures (404,403,forbidden etc) get caught and informed in the logs instead of breaking the plugin with uncaught errors. List: https://onda.qsl.br/temp/URLsWithErrors.txt
Thanks! :)

Rodrigo-NH:
And this is a list with some (471) that i'm using for tests, in case it's useful for you. I could write tests as you suggested but i'm no able to do it ATM as I must check and learn first how to write these.
List --> https://onda.qsl.br/temp/URLsList.txt

plugins/Web/plugin.py
ProgVal: You could make it log all the time, but only raise exceptions if `raiseErrors`.

plugins/Web/plugin.py
ProgVal: Oh ok. That's up to you then

plugins/Web/plugin.py
ProgVal: This line decodes the text, ie turns it from str to bytes, in a lossy way: https://github.com/ProgVal/Limnoria/pull/1377/files#diff-9a20c614ecc345c998e528d2a13f1686R160
And doing `bytes()` on the result does the same thing the other way around.

So it's better to use the original text

ProgVal: Except we're dealing with py2 instead of py3 here, and I forgot how it works. * shrug *

Note that I'll start removing py2 support from Limnoria in 2020

plugins/Web/plugin.py
Rodrigo-NH: I separate these cases <if raiseErrors> goes to 'irc.error' and <if not raiseErrors> goes to logs. I'm not sure about verbosity in line 157 with 'irc.error(_('That URL raised <' + str(e)) + '>',                 Raise=True)' perhaps less verbosity and just informing that 'That URL raised errors'?

plugins/Web/plugin.py
Rodrigo-NH: I just returned to the original value (before my commits)

plugins/Web/plugin.py
Rodrigo-NH: Changed accordingly your suggestion.

Rodrigo-NH:
Tested manually with some good (and bad URLs) in both py3 and py2, working as expected. https://github.com/Rodrigo-NH/Limnoria/commit/f135b0eb67b873bdcfb42d7d29aacf106a9a869b

plugins/Web/plugin.py
```diff
@@ -149,19 +149,31 @@ def noIgnore(self, irc, msg):
...
-                irc.error(_('Could not guess the page\'s encoding. (Try '
-                        'installing python-charade.)'), Raise=True)
-            else:
-                return None
+            if minisix.PY3:
+                if raiseErrors:
+                    irc.error(_('Could not guess the page\'s encoding. (Try '
+                                'installing python-charade.)'), Raise=True)
+                    return None
+                elif not raiseErrors:
```
jlu5: I would condense this into `else:`, since a boolean can only be either True or False.

plugins/Web/plugin.py
```diff
@@ -149,19 +149,31 @@ def noIgnore(self, irc, msg):
...
-        (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
+        try:
+            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
                 timeout=timeout)
+        except Exception as e:
+            if raiseErrors:
+                irc.error(_('That URL raised <' + str(e)) + '>',
+                          Raise=True)
+                return None
+            elif not raiseErrors:
```
jlu5: Same comment as below.

plugins/Web/plugin.py
Rodrigo-NH: Hi jlu5, thanks. Changed for all similar occurrences.

plugins/Web/plugin.py
```diff
@@ -149,19 +149,31 @@ def noIgnore(self, irc, msg):
...
         timeout = self.registryValue('timeout')
-        (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
+        try:
+            (target, text) = utils.web.getUrlTargetAndContent(url, size=size,
                 timeout=timeout)
+        except Exception as e:
+            if raiseErrors:
+                irc.error(_('That URL raised <' + str(e)) + '>',
+                          Raise=True)
+                return None
```
ProgVal: That return is dead code

plugins/Web/plugin.py
```diff
@@ -149,19 +149,31 @@ def noIgnore(self, irc, msg):
...
-            if raiseErrors:
-                irc.error(_('Could not guess the page\'s encoding. (Try '
-                        'installing python-charade.)'), Raise=True)
-            else:
-                return None
+            if minisix.PY3:
+                if raiseErrors:
+                    irc.error(_('Could not guess the page\'s encoding. (Try '
+                                'installing python-charade.)'), Raise=True)
+                    return None
```
ProgVal: same

plugins/Web/plugin.py
Rodrigo-NH: Sorry for delay, been having really busy lately and despite being a simple change I wanted to test it anyway. Done.

ProgVal: you forgot this one

Rodrigo-NH: True. Fixed now.

PR #1379, opened 2019-10-17
regexp_wrapper: return None instead of False when search times out

jlu5:
This allows plugins to detect whether a search timed out or did not match, which are two distinct outcomes.

Currently this affects SedRegex: if the search on one message times out, it assumes that the regex did not match and moves on to the next message. Often times, this ends up causing the wrong message to be replaced.


ProgVal:
Returning `False` is indeed a bad behavior. Raising an exception would probably be better than returning `None`, but it would break existing plugins (although I don't know many that use this feature).
I guess that's okay then

PR #1380, opened 2019-10-18
Remove the /debian directory.

baldurmen:
This is not really a native Debian package and this directory hasn't been touched in years.
The actual Debian packaging work is done on salsa.debian.org.

For more information on native vs non-native packaging, this is a pretty good primer:

https://wiki.debian.org/DebianMentorsFaq#What_is_the_difference_between_a_native_Debian_package_and_a_non-native_package.3F

Cheers!


ProgVal:
Good point. Thanks!

PR #1395, opened 2019-12-15
Web: Add supybot.plugins.Web.requestLanguage

tatokis:
Some websites autodetect the language based on the server location.
This PR adds a variable that lets users override the served language by specifying the `Accept-Language` header.

```
<User> %web title https://twitter.com/
<Bot> Twitter. Ce qu'il se passe.
<User> %config supybot.plugins.Web.requestLanguage "en-GB"
<Bot> Ok
<User> %web title https://twitter.com/
<Bot> Twitter. It's what's happening
```


ProgVal:
Thanks, but I think it should be global to the bot, not just Web.

You could use something similar to https://github.com/ProgVal/Limnoria/blob/1a1707420b58f2cc2223c9fe94f7f6837e08fd8f/src/conf.py#L1278-L1281 to add the header to `utils.web.defaultHeaders`.

PR #1398, opened 2020-01-08
Factoids: Filter duplicates in rank command

tatokis:
If there are multiple factoids with the same key, the rank command will return the key multiple times.
`#1 firstkey (234), #2 secondkey (111), #3 secondkey (111), #4 secondkey (111), ...`

This is corrected by using `SELECT DISTINCT` in the database query.

### DIFF
--- a/plugins/Factoids/plugin.py
+++ b/plugins/Factoids/plugin.py
@@ -565,7 +565,7 @@ class Factoids(callbacks.Plugin, plugins.ChannelDBHandler):
             number = self.registryValue('rankListLength', channel, irc.network)
         db = self.getDb(channel)
         cursor = db.cursor()
-        cursor.execute("""SELECT keys.key, relations.usage_count
+        cursor.execute("""SELECT DISTINCT keys.key, relations.usage_count
                           FROM keys, relations
                           WHERE relations.key_id=keys.id
                           ORDER BY relations.usage_count DESC


ProgVal:
Shouldn't it sum `relations.usage_count` for all keys?

And could you add a unit test, to make sure there's no regression in the future?

Thanks!

tatokis:
> Shouldn't it sum relations.usage_count for all keys?

Good point! However, this brings the following question:

Say you have a factoid key with 3 entries in it.
When you call `key`, it increments the count for all three. If you, however, use `whatis key 2`, then it increments only the count for id 2.

Should we then display the value of the key, say `#2 key (value of id 2), #3 key (rest)`?
Or should we make a list of all the factoids with the same key, find the lowest usage count, subtract it from the rest, and then add the remaining difference to the lowest count, considering that the final value?

This turned out much more complicated than I initially thought it was going to be, as I completely forgot one can display the values separately using the whatis command.

> And could you add a unit test, to make sure there's no regression in the future?

Sure, I'll try to do so after the above is taken care of.

ProgVal:
I don't have a good solution to this problem :/

>  Or should we make a list of all the factoids with the same key, find the lowest usage count, subtract it from the rest, and then add the remaining difference to the lowest count, considering that the final value?

But definitely not that. It would give weird, hard-to-understand results.



I think the least wrong approach would be to blindly sum them all, even if a single call is counted multiple times.

tatokis:
> But not that. It would give weird, hard-to-understand results.

I actually thought that was the better solution, since it would count how many times the key itself was called, as from what I understand, that is the point of the `rank` command.
`Returns a list of top-ranked factoid keys, sorted by usage count (rank).`

By finding the minimum, you can count how many times all the factoids in the key were shown at least once, and then by adding the difference from the minimum for each key to the total, you count how many times each remaining key was called individually.

In pseudocode
```
total = m = min(dupkeys)
for k in dupkeys:
    total += k.count - m
```

From what I understand, It isn't possible to call a key and only show a specified number of factoids. It's either one or all, so this method should be pretty accurate in counting how many times the key was called.
Unless of course I am missing something in the thought process, which may be very likely.

That said, I don't like the idea of blindly summing them, as that can give extremely inaccurate results.

For example, if I have a factoid key with 5 values and they only get called together, as is the case with many of them in my bot, then the count number will eventually be very high and not represent the number of times `key` was called.
If someone calls the factoid key 5 times (displaying all 5 factoids), then the total will be 25, which I think is even more confusing, as they obviously only called `key` 5 times.

Assuming it's not much of a maintenance trouble, I wouldn't be against making this configurable and including multiple 'algorithms'.
Otherwise, I think the other solution I suggested (displaying the factoids in the rank command) would be a decent compromise.

ProgVal:
> > But not that. It would give weird, hard-to-understand results.
> 
> I actually thought that was the better solution, since it would count how many times the key itself was called, as from what I understand, that is the point of the `rank` command.
> `Returns a list of top-ranked factoid keys, sorted by usage count (rank).`
> 
> By finding the minimum, you can count how many times all the factoids in the key were shown at least once, and then by adding the difference from the minimum for each key to the total, you count how many times each remaining key was called individually.
> 
> In pseudocode
> 
> ```
> total = m = min(dupkeys)
> for k in dupkeys:
>     total += k.count - m
> ```
> 
> From what I understand, It isn't possible to call a key and only show a specified number of factoids. It's either one or all, so this method should be pretty accurate in counting how many times the key was called.
> Unless of course I am missing something in the thought process, which may be very likely.

If there are 5 dup keys, and each one was shown once, this algo will show the number of calls to be 1 instead of 5.

> 
> That said, I don't like the idea of blindly summing them, as that can give extremely inaccurate results.
> 
> For example, if I have a factoid key with 5 values and they only get called together, as is the case with many of them in my bot, then the count number will eventually be very high and not represent the number of times `key` was called.
> If someone calls the factoid key 5 times (displaying all 5 factoids), then the total will be 25, which I think is even more confusing, as they obviously only called `key` 5 times.

yeah :/

> 
> Assuming it's not much of a maintenance trouble, I wouldn't be against making this configurable and including multiple 'algorithms'.

Let's try this and see how complex the code gets

PR #1399, opened 2020-01-12
Add supybot.protocols.http.requestLanguage

tatokis:
Some websites autodetect the language based on the server location.
This PR adds a global variable that lets users override the served language by specifying the `Accept-Language` header.
The Web plugin has also been modified to implement channel-specific language settings.

```
<User> %web title https://twitter.com/
<Bot> Twitter. Ce qu'il se passe.
<User> %config supybot.protocols.http.requestLanguage "en-GB"
<Bot> Ok
<User> %web title https://twitter.com/
<Bot> Twitter. It's what's happening
```

Continuation of https://github.com/ProgVal/Limnoria/pull/1395

### DIFF
--- a/plugins/Web/plugin.py
+++ b/plugins/Web/plugin.py
@@ -146,12 +146,20 @@ class Web(callbacks.PluginRegexp):
     def noIgnore(self, irc, msg):
         return not self.registryValue('checkIgnored', msg.channel, irc.network)
 
-    def getTitle(self, irc, url, raiseErrors):
+    def getTitle(self, irc, url, raiseErrors, msg):
         size = conf.supybot.protocols.http.peekSize()
         timeout = self.registryValue('timeout')
+        headers = dict(utils.web.defaultHeaders)
+        language = conf.supybot.protocols.http.requestLanguage.getSpecific(
+                irc.network, msg.channel)()
+        lkey = 'Accept-Language'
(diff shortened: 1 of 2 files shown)


plugins/Web/plugin.py
```diff
@@ -146,12 +146,17 @@ class Web(callbacks.PluginRegexp):
...
 
-    def getTitle(self, irc, url, raiseErrors):
+    def getTitle(self, irc, url, raiseErrors, msg):
         size = conf.supybot.protocols.http.peekSize()
         timeout = self.registryValue('timeout')
+        headers = dict(utils.web.defaultHeaders)
+        language = conf.supybot.protocols.http.requestLanguage.getSpecific(
+                irc.network, msg.channel)()
+        if language:
+            headers['Accept-Language'] = language
```
ProgVal:
```
elif not language:
    del headers['Accept-Language']
```

tatokis: If we add that, then the global value will not be inherited, as it will just remove the inherited header from the dict altogether.

ProgVal: The channel-specific value of `supybot.protocols.http.requestLanguage` inherits the global value already. So if it's empty, it means that either the global value is empty, or it was set to empty specifically for the channel.

PR #1401, opened 2020-02-01
Missing German translations

EvilOlaf:
Added a few missing translations in Deutsch.
They are not perfect. Better than having nothing I guess  ¯\_(ツ)_/¯ 

I may add a few more in future if I find some free time.


ProgVal:
Thanks!

PR #1404, opened 2020-02-18
Relay: Add `nonPrivmsgs` config option

elieux:
Can be used to disable relaying of non-PRIVMSG messages, since in a quiet channel it generates a lot of potentially unwanted messages/notices.

Please comment.


ProgVal:
I would rather have `nonPrivmsg` take three values, like LinkRelay does: https://github.com/ProgVal/Supybot-plugins/blob/master/LinkRelay/config.py#L81-L87
and then mark `noticeNonPrivmsgs` as deprecated (but keep it working for old bots).

Could you do that?

Thanks!

elieux:
Okay, sounds good.

elieux:
As discussed on IRC, this old plugin should be replaced by LinkRelay or RelayNext, rather than extended with features.

Issue #1372, opened 2019-10-05
Herald: config parameters do not applying to default herald


koobs:
As investigated with @ProgVal on freenode, _at least_ Herald config's `throttle` and `throttle.afterPart` don't apply to the **default** herald, when they should. It's probable that `afterSplit` and other configs are also affected.

Current behaviour:

Cycling the channel, after setting `throttle` or `afterPart`, results in herald messaging (heralding) the user multiple times.

Expected behaviour:

Herald does **not** send the default herald message to a user if/when  `throttle` or `throttle.afterPart` (or `afterSplit`) are set/matched for that user.

lsalvadore:
Just a guess.

I think the problem is in line 119 and line 123 of Herald/plugin.py: ` if now - self.lastHerald.get((channel, id), 0) > throttle:` and `if now - self.lastParts[channel, id] < i:`.
You calculate the differences `now - self.lastHerald.get((channel, id), 0)` and `now - self.lastParts[channel, id]`, which, when a user cycles the channel, are extremely small values: extremely small values often mess up codes.
In your case these extremely small values are then compared with `throttle` and `throttle.afterPart`, which are defined in config.py as `PositiveInteger` and `NonNegativeInteger` respectively. I do not know how types and conversions are implemented in Limnoria, but I guess the extremely small value is converted to PositiveInteger and NonNegativeInteger in such a way that it becomes an extremely high value. Do you think it is possible?

If conversion of extremely small values is indeed the problem, explicit conversion would probably do the trick, i.e. substituing the lines with `if int(now - self.lastHerald.get((channel, id), 0)) > throttle:` and `if int(now - self.lastParts[channel, id]) < i:`.

lsalvadore:
It has been pointed out that the bug concerns the default hearld only, not all heralds (I missed that part in the OP), so my guess is probably wrong.
I am sorry, I will try to study better the plugin.

koobs:
@ProgVal Hey mate :) We've been blocked on this issue since reporting, to making substantial changes to our bot for #freebsd* (> 50) channels on freenode. I'm trying to find someone in the community to resolve this, but if you have spare cycles or are otherwise able to prioritize this one, provide guidance on resolution (so we can more easily submit a PR) or know of people in the Limnoria community who might be able to assist, that would be greatly apprecisted

Freaky:
Those `time.time()` calls look like they should be `time.monotonic()`, so throttling isn't broken by system clock changes.

[...]

PR #1419, opened 2020-07-03
Herald: Throttle default heralds.

oddluck:
Fixes https://github.com/ProgVal/Limnoria/issues/1372

### DIFF
--- a/plugins/Herald/plugin.py
+++ b/plugins/Herald/plugin.py
@@ -87,31 +87,15 @@ class Herald(callbacks.Plugin):
     def doJoin(self, irc, msg):
         if ircutils.strEqual(irc.nick, msg.nick):
             return # It's us.
-        if msg.nick in self.splitters:
-            self.log.debug('Not heralding %s, recent split.', msg.nick)
-            return # Recently split.
         channel = msg.args[0]
         irc = callbacks.SimpleProxy(irc, msg)
         if self.registryValue('heralding', channel, irc.network):
             try:
                 id = ircdb.users.getUserId(msg.prefix)
-                if id in self.splitters:
-                    self.log.debug('Not heralding id #%s, recent split.', id)
-                    return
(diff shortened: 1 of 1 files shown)


plugins/Herald/plugin.py
```diff
@@ -87,31 +87,15 @@ def doQuit(self, irc, msg):
...
-                        msgmaker = ircmsgs.notice
-                    target = msg.nick
-                    if self.registryValue('default.public',
-                                          channel, irc.network):
-                        target = channel
-                    irc.queueMsg(msgmaker(target, default))
+                id = msg.nick
+            if id in self.splitters:
+                self.log.debug('Not heralding id #%s, recent split.', id)
                 return
```
ProgVal, line 99: this change (using `id in self.splitters` instead of `msg.nick in self.splitters`) is not related to throttling, could you move it to a new PR?

plugins/Herald/plugin.py
```diff
@@ -122,16 +106,34 @@ def doJoin(self, irc, msg):
...
+                    irc.reply(herald, prefixNick=False)
 
     def doPart(self, irc, msg):
         try:
             id = self._getId(irc, msg.prefix)
-            self.lastParts[msg.args[0], id] = time.time()
         except KeyError:
-            pass
+            id = msg.nick
+        self.lastParts[msg.args[0], id] = time.time()
```
ProgVal, line 136: same

plugins/Herald/plugin.py
oddluck: the reason throttle.afterpart doesn't work is because it only records parts for registered users, the fix only works because it now records msg.nick, so moving it to a new pull request is not really practical...

ProgVal: Oh, I missed we already have `id in self.splitters`. nvm then

PR #1420, opened 2020-07-06
DDG: Fix redirects, add region/safeSearch configs.

oddluck:
DDG redirect detection is currently broken, the first commit fixes that. DDG also uses the host IP to auto-detect region, the second commit adds a region config to allow setting the language/country of your choice. Also in the second commit is support for Safe Search, bringing DDG more in line with the Google plugin.


ProgVal:
Could you add a test for the region setting?

Issue #1382, opened 2019-10-21
Make the user-agent for HTTP requests configurable


ProgVal:
May be useful to access websites with rules blocking the default user-agent.

oddluck:
I could tackle this one, do you want to just use one user-agent or randomly select from a list like SpiffyTitles?

koobs:
@oddluck Thanks for offering to pick this up Gordan! :)

From our use of Limnoria for #freebsd* channel management on freenode IRC, i can summarize our needs with respect to User-Agent as follows:

 1. Set a default and global User-Agent value that applies to all outbound requests

This allows all requests to have a safe/unique default. We use this so our Plugins that use *.freebsd.org services can uniquely identify themselves, to ascertain IRC service usage numbers.

 1. Set or override User-Agent value on a per-plugin basis 

This allows a plugin to set a target specific user-agent. Example: A TweetPlugin can set User-Agent: X needed only for interacting with the Twitter API.

 1. Set or override User-Agent on an individual request or plugin method/class basis

This allows a plugin with multiple target/destinations to set class specific user-agent value. Example: a NewsPlugin with multiple news distinct sources that might require custom User-Agent values specific to it

oddluck:
@koobs have you looked at [SpiffyTitles](https://github.com/oddluck/limnoria-plugins/tree/master/SpiffyTitles)? It already handles user agents per-channel as well as offering different handlers for Twitter links etc. and is better equipped to handle such changes than limnoria core. I was just looking to close this issue not rewrite the web plugin when SpiffyTitles already covers the same ground and is a recommended plugin at https://limnoria.net/plugins.xhtml

koobs:
@oddluck Thanks for the prompt response Gordon, and yep we're aware of spiffytitles :)

The originating motivation that prompted @ProgVal to kindly created this issue was for requests that Limnoria can make for plugins not related to title snarfing.

This feels very much in 'core framework' territory, and I would expect that SpiffyTitles (and other plugins) [1] would/could use that core framework mechanism for its requests once landed.

[1] https://github.com/oddluck/limnoria-plugins/search?q=user-agent&unscoped_q=user-agent
[2] https://github.com/oddluck/limnoria-plugins/blob/master/SpiffyTitles/plugin.py

koobs:
One might start or increment our way there by having a / moving the get_user_agent() method to core, where if its a single entry, its static (configurable), and if it contains multiple entries a random one is returned. 

[...]

PR #1426, opened 2020-07-14
Google: Set User-Agent override.

oddluck:
Fixes https://github.com/ProgVal/Limnoria/issues/1382#issuecomment-654617022. This also provides case consistency (User-Agent vs. User-agent) across all plugins... See https://github.com/ProgVal/Limnoria/blob/138957676ec8c974dde1e45be81a37691f17434c/src/utils/web.py#L120 for further confirmation...

### DIFF
--- a/plugins/Google/plugin.py
+++ b/plugins/Google/plugin.py
@@ -107,6 +107,7 @@ class Google(callbacks.PluginRegexp):
                                     dynamic.irc.nick)
         headers = dict(utils.web.defaultHeaders)
         headers['Referer'] = ref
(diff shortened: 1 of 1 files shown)


plugins/Google/plugin.py
```diff
@@ -254,7 +255,7 @@ def fight(self, irc, msg, args):
 
     def _translate(self, sourceLang, targetLang, text):
         headers = dict(utils.web.defaultHeaders)
```
ProgVal, line 257: With `.copy()` instead of `dict()`, case shouldn't matter on the other line

PR #1431, opened 2020-10-03
Implement text reminders

mogad0n:
Implemented text based reminders. Appropriately changed _restoreEvents()

TODO: `@scheduler list` should not display reminders as they may contain sensitive data.

[flake8 (http link) ](http://ix.io/2zzU) 

L84 may be redundant

tests

### DIFF
--- a/plugins/Scheduler/plugin.py
+++ b/plugins/Scheduler/plugin.py
@@ -90,7 +90,9 @@ class Scheduler(callbacks.Plugin):
                         # though we'll never know for sure, because other
                         # plugins can schedule stuff, too.
                         n = int(name)
-                    self._add(network, event['msg'], event['time'], event['command'], n)
+                    isreminder = event.get('isreminder')
+                    self._add(irc, event['msg'], event['time'], event['command'],
+                              isreminder, n)
                 elif event['type'] == 'repeat': # repeating event
                     now = time.time()
                     first_run = event.get('first_run')
@@ -105,7 +107,7 @@ class Scheduler(callbacks.Plugin):
                     next_run_in = self._getNextRunIn(
                         first_run, now, event['time'], not_right_now=True)
 
-                    self._repeat(network, event['msg'], name,
+                    self._repeat(irc, event['msg'], name,
                                  event['time'], event['command'], first_run, next_run_in)
             except AssertionError as e:
                 if str(e) == 'An event with the same name has already been scheduled.':
@@ -133,11 +135,9 @@ class Scheduler(callbacks.Plugin):
         world.flushers.remove(self._flush)
         self.__parent.die()
 
-    def _makeCommandFunction(self, network, msg, command, remove=True):
+    def _makeCommandFunction(self, irc, msg, command, remove=True):
         """Makes a function suitable for scheduling from command."""
         def f():
-            # If the network isn't available, pick any other one
-            irc = world.getIrc(network) or world.ircs[0]
             tokens = callbacks.tokenize(command,
                 channel=msg.channel, network=irc.network)
             if remove:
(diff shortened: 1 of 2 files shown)


plugins/Scheduler/plugin.py
```diff
@@ -90,7 +90,9 @@ def _restoreEvents(self, irc):
                         # though we'll never know for sure, because other
                         # plugins can schedule stuff, too.
                         n = int(name)
-                    self._add(network, event['msg'], event['time'], event['command'], n)
+                    isreminder = event.get('isreminder')
```
ProgVal, line 93: It's equivalent, but I'd like to use `isreminder = event.get('isreminder', False)`, so a missing value returns `False` instead of `None`.

Also, please add a comment explaining that we need `.get()` instead of `[]` older bots didn't have it set

plugins/Scheduler/plugin.py
```diff
@@ -133,25 +135,34 @@ def die(self):
...
                 del self.events[str(f.eventId)]
             self.Proxy(irc, msg, tokens)
         return f
 
-    def _add(self, network, msg, t, command, name=None):
-        f = self._makeCommandFunction(network, msg, command)
+    def _makeReminderFunction(self, irc, msg, text):
+        """Makes a function suitable for scheduling text"""
+        def f():
+            irc.reply('Reminder: %s' % text, msg=msg, prefixNick=True)
```
ProgVal, line 151: `_('Reminder: %s') % text`

plugins/Scheduler/plugin.py
```diff
@@ -167,10 +178,24 @@ def add(self, irc, msg, args, seconds, command):
...
+    def remind(self, irc, msg, args, seconds, text):
+        """ <seconds> <text>
+
+        Sets a reminder with string <text> to run <seconds> seconds in the
+        future. For example, 'scheduler remind [seconds 30m] "Hello World"'
+        will return '<nick> Reminder: Hello World' 30 minutes after being set.
+        """
+        t = time.time() + seconds
+        isreminder = True
+        id = self._add(irc, msg, t, text, isreminder)
```
ProgVal, line 195: `id = self._add(irc, msg, t, text, is_reminder=True)`

plugins/Scheduler/test.py
```diff
@@ -74,6 +74,13 @@ def test00RemoveZero(self):
         timeFastForward(5)
         self.assertNoResponse(' ', timeout=1)
 
+    def testRemind(self):
+        self.assertNotError('scheduler remind 5 testRemind')
+        timeFastForward(5)
+        self.assertResponse(' ', 'Reminder: testRemind')
+        timeFastForward(3)
```
ProgVal, line 81: remove this one, not needed.

plugins/Scheduler/test.py
```diff
@@ -74,6 +74,13 @@ def test00RemoveZero(self):
         timeFastForward(5)
         self.assertNoResponse(' ', timeout=1)
 
+    def testRemind(self):
+        self.assertNotError('scheduler remind 5 testRemind')
+        timeFastForward(5)
+        self.assertResponse(' ', 'Reminder: testRemind')
```
ProgVal, line 80: This doesn't make sure the reminder isn't sent too early.

Instead you should use `timeFastForward` on a shorter duration, use `assertNoResponse` to check the reminder isn't set yet, `timeFastForward` again, then `assertResponse`.

ProgVal (changes requested):

ProgVal:
Could you also edit your commit message to explain why this is useful (because we want to allow text reminders without allowing arbitrary commands, etc.), so future readers can understand why you implemented it?

Also, could you rename `isreminder` to `is_reminder` throughout your code?

plugins/Scheduler/plugin.py
```diff
@@ -90,7 +90,12 @@ def _restoreEvents(self, irc):
                         # though we'll never know for sure, because other
                         # plugins can schedule stuff, too.
                         n = int(name)
-                    self._add(network, event['msg'], event['time'], event['command'], n)
+                    is_reminder = event.get('is_reminder', False)
+                    # Here we use event.get() method instead of event[]
+                    # This is to maintain compatibility with older bots
+                    # lacking 'is_reminder' in their event dicts
```
ProgVal, line 96: swap the comment and the definition of `is_reminder` (it's more common to write comments before the code they are describing)

plugins/Scheduler/test.py
```diff
@@ -74,6 +74,13 @@ def test00RemoveZero(self):
         timeFastForward(5)
         self.assertNoResponse(' ', timeout=1)
 
+    def testRemind(self):
+        self.assertNotError('scheduler remind 5 testRemind')
+        timeFastForward(3)
+        self.assertNoResponse(' ', timeout=1)
+        timeFastForward(2)
+        self.assertResponse(' ', 'Reminder: testRemind')
```
ProgVal, line 82: you should re-add the `timeFastForward(5)` + `assertNoResponse` at the end to check it's not repeated.

plugins/Scheduler/test.py
```diff
@@ -74,6 +74,13 @@ def test00RemoveZero(self):
         timeFastForward(5)
         self.assertNoResponse(' ', timeout=1)
 
+    def testRemind(self):
+        self.assertNotError('scheduler remind 5 testRemind')
+        timeFastForward(3)
```
ProgVal, line 79: Make it 4; because 3+2 is too close to 5, so `assertResponse` might run just before it's scheduled. With 4+2 it's sure it will run after the reminder is scheduled

ProgVal (changes requested):
and please undo the docstring changes (and configure your editor so it stops doing that)

PR #1432, opened 2020-10-03
add unrestricted text reminder scheduling commands as scheduler.add can be subject to abuse 

mogad0n:
#1431


ProgVal:
Great! Now, can you add a test for `list` when there are reminders?

mogad0n:
I believe that implementation is proper

ProgVal:
just one more at the end of the test to make sure it's not shown anymore

PR #1435, opened 2020-10-29
resolve warnings about 'is not' and integer comparisons

larsks:
running supybot-adduser (or supybot-reset-password) with a recent
python would result in the warning:

> ```
> /usr/local/bin/supybot-adduser:59: SyntaxWarning: "is not" with a literal. Did you mean "!="?
> if len(args) is not 1:
> ```

This commits corrects the syntax.
`


jlu5:
Thanks!

PR #1437, opened 2020-12-10
Fix HTML entities not escaped in RSS feed title

EdmundOgban:
RSS announce does not currently unescape HTML entities.

I scratched my head on where exactly to unescape them, and doing that when announcing seems the more reasonable place to do it.
I also looked if feedparser could already do that, but to no avail.

Before:
> testfeed: Generic support • Re: Channels do not remain &amp;quot;persistent&amp;quot;

After:
> testfeed: Generic support • Re: Channels do not remain "persistent"

### DIFF
--- a/plugins/RSS/plugin.py
+++ b/plugins/RSS/plugin.py
@@ -39,6 +39,7 @@ import string
 import socket
 import threading
 import feedparser
+import html
 
 import supybot.conf as conf
 import supybot.utils as utils
@@ -430,6 +431,7 @@ class RSS(callbacks.Plugin):
                     self.registryValue(key_name, channel, network)
         else:
             template = self.registryValue(key_name, channel, network)
+        entry['title'] = html.unescape(entry['title'])
         date = entry.get('published_parsed')
         date = utils.str.timestamp(date)
         s = string.Template(template).substitute(


ProgVal:
Hi,

Feedparser should already escape those (the XML parser that Feedparser runs, actually).

For example, this feed:

```
<?xml version="1.0" encoding="utf-8"?>
<rss version="2.0">
<channel>
<item>
<title>test &amp; test</title>
<link>http://example.org/entry/3</link>
</item>
</channel>
</rss>
```

is parsed like this:

```
>>> pprint(feedparser.parse("./foo.rss"))
{'bozo': 0,
 'encoding': 'utf-8',
 'entries': [{'link': 'http://example.org/entry/3',
              'links': [{'href': 'http://example.org/entry/3',
                         'rel': 'alternate',
                         'type': 'text/html'}],
              'title': 'test & test',
              'title_detail': {'base': '',
                               'language': None,
                               'type': 'text/plain',
                               'value': 'test & test'}}],
 'feed': {},
 'namespaces': {},
 'version': 'rss20'}
```

Do you have an example of a feed where this isn't true?

EdmundOgban:
Hm, that's curious.
Of course I have: https://forums.unrealircd.org/viewtopic.php?t=9058&p=39735#p39735

EdmundOgban:
Ok, so, I was wondering if it was because it's in a CDATA block, and that seems to be the problem

ProgVal:
You're talking about https://forums.unrealircd.org/app.php/feed?sid=160db7297b0b946912a2c2cf24c6cde1 , the RSS feed linked from that page, right?

The only issue I see when parsing that feed is:

```
'title': 'Generic support • Re: Channels do not remain &quot;persistent&quot; on unrealircd 5.0.6'
```

But this is coming from this part of the RSS feed:

```
<title type="html"><![CDATA[Generic support • Re: Channels do not remain &quot;persistent&quot; on unrealircd 5.0.6]]></title>
```

And `<![CDATA[` explicitly says what is inside should not be escaped.

So if I am not mistaken, the bug is in the RSS feed generator.

EdmundOgban:
Yes, that makes sense but, while technically correct, it seems strange to me that the generator is doing something that wrong. I'm assuming that, since it already happened, it would happen again. I'd prefer to have a nicer output than be extremely compliant with the specs. After all, RSS's job is to present text for human consumption, no?

ProgVal:
> it seems strange to me that the generator is doing something that wrong

That's not unique at all. I get a couple reports every year about broken RSS feeds, some in ways much worse than that (eg. not even valid XML)

> it would happen again. I'd prefer to have a nicer output than be extremely compliant with the specs

I'm not against it on principle, there are already a couple of workaround for broken feeds in the RSS code; and feedparser itself also does a lot of fixing for broken feeds when it can.

But there are legitimate reasons to have `&quote;` in the real title for example; and this change would break it. Yes, it's unlikely, but I would rather not break 1 legitimate feed in order to fix even 10 broken ones.

EdmundOgban:
Fair enough. Anyway, for the record, I've tried an RSS reader extension for Chrome and actually does the unescape.
I think that the real problem is in case you're evaluating the content of the CDATA block, but in this case we're still treating it as text. I'm still +1 for the change; I'm leaving it in my fork and keeping an eye on it and maybe report back, if it can help.

ProgVal:
Did you report this to unrealircd.org / phpbb? They should fix this on their side either way

EdmundOgban:
No, but I don't think they will care much. It seems that not applying any trasformation to the content of the CDATA block is a frequently broken rule...

ProgVal:
I'm closing this, because I don't think that's an issue should fix.

PR #1449, opened 2021-04-24
Per network services

elieux:
Please review, partially discussed on IRC.  Since there's a bunch of fixes, I recommend going commit by commit.

### DIFF
--- a/plugins/Services/plugin.py
+++ b/plugins/Services/plugin.py
@@ -42,6 +42,13 @@ import supybot.callbacks as callbacks
 from supybot.i18n import PluginInternationalization, internationalizeDocstring
 _ = PluginInternationalization('Services')
 
+class State:
+    def __init__(self):
+        self.channels = []
+        self.sentGhost = None
+        self.identified = False
+        self.waitingJoins = []
+
 class Services(callbacks.Plugin):
     """This plugin handles dealing with Services on networks that provide them.
     Basically, you should use the "password" command to tell the bot a nick to
@@ -66,10 +73,10 @@ class Services(callbacks.Plugin):
         self.reset()
 
     def reset(self):
-        self.channels = []
-        self.sentGhost = None
-        self.identified = False
-        self.waitingJoins = {}
+        self.state = {}
+
+    def _getState(self, irc):
+        return self.state.setdefault(irc.network, State())
 
     def disabled(self, irc):
         disabled = self.registryValue('disabledNetworks')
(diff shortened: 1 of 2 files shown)


plugins/Services/plugin.py
```diff
@@ -66,10 +66,16 @@ def __init__(self, irc):
...
-        self.waitingJoins = {}
+        self.state = {}
+
+    def _getState(self, irc):
+        class State(object):
+            def __init__(self):
+                self.channels = []
+                self.sentGhost = None
+                self.identified = False
+                self.waitingJoins = []
```
ProgVal, line 77: Let's not create a class on every call... define it at the toplevel.
And no need to inherit from `object`

plugins/Admin/test.py
```diff
@@ -31,13 +31,17 @@
...
     def testChannels(self):
         def getAfterJoinMessages():
             m = self.irc.takeMsg()
+            self.assertIsNotNone(m, 'No message back from joining channel.')
             self.assertEqual(m.command, 'MODE')
             m = self.irc.takeMsg()
+            self.assertIsNotNone(m, 'No message back from joining channel.')
             self.assertEqual(m.command, 'MODE')
             m = self.irc.takeMsg()
+            self.assertIsNotNone(m, 'No message back from joining channel.')
```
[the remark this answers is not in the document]
elieux: What about this then? https://github.com/ProgVal/Limnoria/blob/testing/src/test.py#L492-L499

plugins/Services/test.py
```diff
@@ -105,6 +106,69 @@ def testRegisterNoExperimentalExtensions(self):
...
+            self.assertEquals(self.irc.takeMsg(), queuedJoin, 'Join request did not go through after identification.')
+        finally:
+            self.assertNotError('services password %s ""' % self.nick)
+
+    def testMultipleNetworks(self):
+        try:
+            net1 = copy(self)
+            net1.irc = getTestIrc('testnet1')
+            net1.assertNotError('services password %s %s' % (self.nick, 'secret'))
+            try:
```
ProgVal, line 152: You can probably do this with a single `try` block

ProgVal (changes requested):
`assertEqual` and 80 cols, please

plugins/Admin/test.py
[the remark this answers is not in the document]
ProgVal: Uh, I don't know. But most plugins don't check it.

plugins/Services/plugin.py
elieux: Okay and okay.

plugins/Services/test.py
elieux: How to make sure both of the finally statements are attempted though?

plugins/Admin/test.py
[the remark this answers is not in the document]
elieux: Commit removed.

plugins/Services/plugin.py
elieux: Done and done.

plugins/Services/test.py
elieux: I guess the strictly correct behavior isn't worth the decrease in readability.  I merged the two blocks.

elieux:
> `assertEqual`

I didn't know.  Changed now.

> 80 cols, please

Sorry.  I think I managed to fix all the long lines tastefully.

PR #1458, opened 2021-05-27
Config: note "config flush false" when option is not writeable via IRC

Mikaela:


ProgVal:
Why? That shouldn't be needed unless it takes more than an hour to edit the config and reload it

Mikaela:
I just noticed it was missing and thought that it would be a good thing to document so there is no config getting written and possibly causing conflict/confusion with text editor simultaneously.

Isn't more than an hour a bit long time between flushes if something changes?

PR #1464, opened 2021-05-31
setup.py: switch import to setuptools

ianw:
Switch to standard setuptools import, add suggested entries to
pyproject.toml.

Remove the --clean argument.  As the comment suggests I'm sure there
is history here, but having setup.py remove parts of the package does
not seem like something required at this point.

Also clean up the imports to remove unused and group them together at
the top.


ProgVal:
Thanks, that's something I wanted to do for a while. I'll just need to make sure it doesn't break existing installs.

ianw:
Note we are using this in production with a python 3.9 container @ https://opendev.org/opendev/system-config/src/branch/master/docker/ircbot

ProgVal:
hmm, I kind of forgot about this. I'll try to look into it next week

setup.py
```diff
@@ -31,12 +31,12 @@
...
 import os
+import subprocess
 import sys
 import time
 import warnings
-import datetime
-import tempfile
-import subprocess
+
+from setuptools import setup
```
ProgVal, line 39: Catch `ImportError` and explain how to install setuptools if it isn't available. (Recommend installing it with the package-manager first, but also explain how to install pip and run `pip3 install --upgrade setuptools` in case setuptools isn't available in the package-manager).

See the deleted distutils comment for examples of how to present it.

pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools", "wheel"]
```
ProgVal, line 2: I don't think we need/should require wheels, falling back to sdist is perfectly fine.
(Yeah I know it's the default snippet, but I don't like adding dependencies we don't need)

ProgVal (changes requested):
Sorry, I forgot to review this, again. I'm fine with the PR.

The JOIN commit was included in this diff, could you remove it?

Thanks!

setup.py
ianw: I think this is best left to pyproject.toml which specifies setuptools as the build dependency.  Otherwise this is just giving probably incorrect instructions -- are you installing globally?  are you in a virtualenv or venv? (virtualenv at least automatically installs setuptools so something else is then wrong)  are you even using pip?

ProgVal: Existing users are used to running `setup.py`, which doesn't read `pyproject.toml` if setuptools isn't installed.

>  Otherwise this is just giving probably incorrect instructions

Then the script should say they must install setuptools somehow, and give some hints/links to do it.

pyproject.toml
ianw: Again, I don't think it's worth deviating from the standard here.  As you note setuptools tells you do this [1].  I suspect (although I'm not certain), it's so "pip wheel" just works and doesn't fail with odd errors.  For example our docker build process has a "builder" phase that produces wheels in the base environment, that are then installed into our production images so they don't have build dependencies.  This is just to say it's not an odd or specific thing to be making wheels these days.  I think the project will benefit more from matching the status quo.

[1] https://setuptools.readthedocs.io/en/latest/build_meta.html

ProgVal: This is a very recent new addition, their documentation didn't even mention it a year ago: https://web.archive.org/web/20200720013741/https://setuptools.readthedocs.io/en/latest/setuptools.html

setuptools will be fine without it.

setup.py
ianw: Please feel free to edit this pull request, as I think you have something specific in mind that is probably best not filtered through me as well.

pyproject.toml
ianw: It seems to me that PEP 517 requires a build_wheel entry point, so a compliant [build-system] section needs to be able to implement that.  I imagine that is why it is in there.  It would probably be good to clarify this

ProgVal: I see. Can we get rid of `[build-system]` entirely and just use `setup.py` like we used to before PEP 517, then?

ianw: The project already has a ```pyproject.toml``` but without ```build-system``` and so, per [1], pip is already assuming it uses setuptools/wheel

> Projects with a pyproject.toml file will use a PEP 517 backend. Projects with a pyproject.toml file, but which don’t have a build-system section, will be assumed to have the following backend settings:
> [build-system]
> requires = ["setuptools>=40.8.0", "wheel"]
> build-backend = "setuptools.build_meta:__legacy__"

All this is doing is being explicit about it.

[1] https://pip.pypa.io/en/stable/cli/pip/#pep-517-and-518-support

ProgVal: ouch, ok then

ProgVal:
(I'm merging this despite the test failures because I don't see any reason why this PR would break tests, so I assume it's just Github's CI being weird)

Anyway, thanks, I'll polish it in the `testing` branch

PR #1473, opened 2021-06-14
Fix joins to many channels

cboylan:
If you have enough channels that the 512 byte message limit on the JOIN
message is hit then limnoria was losing the channel that put it over the
limit and not including it in the next JOIN message. This resulted in
losing one channel for every JOIN message that pushed us over 512 bytes.

We fix this by generating the JOIN message immediately after resetting
the channels list to ensure we include the channel that pushed us over
the limit. Then the next time through our JOIN msg construction we'll
add subsequent channels without forgetting the one that pushed us over.

### DIFF
--- a/test/test_conf.py
+++ b/test/test_conf.py
@@ -1,0 +1,59 @@
+##
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#   * Redistributions of source code must retain the above copyright notice,
+#     this list of conditions, and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright notice,
+#     this list of conditions, and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#   * Neither the name of the author of this software nor the name of
+#     contributors to this software may be used to endorse or promote products
+#     derived from this software without specific prior written consent.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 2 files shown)


ProgVal:
Oops. could you add a regression test in `tests/test_conf.py`?

fungi:
Thanks for pushing this. I've applied it manually in our deployment and it's corrected the behavior. We configure a bot to join 88 channels presently, and this was getting split into multiple messages causing us to skip joining a random 2 channels at the boundaries between the messages. Restarting with this applied we joined all 88 correctly.

test/test_conf.py
```diff
@@ -0,0 +1,59 @@
...
+            name = "#verylongchannelname" + str(x)
+            channels.add(name)
+            input_list.append(name)
+        conf.supybot.networks.test.channels.setValue(channels)
+        msgs = conf.supybot.networks.test.channels.joins()
+        # Double check we split the messages
+        self.assertEqual(len(msgs), 2)
+        # Ensure all channel names are present
+        chan_list = (msgs[0].args[0] + ',' + msgs[1].args[0]).split(',')
+        self.assertEqual(sorted(input_list), sorted(chan_list))
```
ProgVal, line 58: `         self.assertCountEqual(input_list, chan_list)

ProgVal:
what did you change?

cboylan:
I added the requested `self.assertCountEqual(input_list, chan_list)`.

ProgVal:
Oh I missed it. I meant it as a replacement for the comparison between sorted()

cboylan:
I can make that change, though it seems that comparing the actual channel names given as input against the channels listed in the output is also a good check? The count would catch this particular issue, but it is possible some other transformation could happen that you would want to guard against?

ProgVal:
`assertCountEqual` does compare the values, not just the lengths: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

cboylan:
> `assertCountEqual` does compare the values, not just the lengths: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

Aha. Today I learned. Commit has been updated to do only assertCountEqual().

ProgVal:
Thanks! :)

Issue #1495, opened 2021-11-13
Karma upvotes for users/nicks; fixing the delimiters 


prdes:
According to the current setup, when the following  `nick` patterns are upvoted (++),

1. `nick++`
2. `nick ++`
3. `nick: ++`
4. `nick, ++`

The increment doesn't strip the space char, nor the various other chars that clients seem to have appended to autocompleted nicks as defaults, resulting in every pattern having 1 vote.

mogad0n:
Now this is the concept borrowed from the karma implementation in `Bitbot` in terms of the regex and the reverse karma (which isn't particularly necessary imo) but the parenthesis version is interesting.

Issue is that it is only using `++` and `--` as the inc and dec chars respectively.

P.S. Naturally there are tests to be written

mogad0n:
I'll just use `%s` for the inc and dec variables.

progval:
Solved by 01cdfee53e9ab4858c1ffff149e5a43510c4fa1a

PR #1496, opened 2021-11-16
Karma: Strip Spaces and characters appended to Nicks by clients

mogad0n:
Attempt towards #1495 

Issues:
Hardcoded inc/dec chars


mogad0n:
I'm unconvinced and shall test on instance and write tests. How `rlstrip()` works that way!

mogad0n:
The squash and fixup didnt go perfect. The space, `,` and `:` chars are `rstrip()`'d from nicks

Added tests for the same

plugins/Karma/plugin.py
```diff
@@ -315,7 +315,10 @@ def doPrivmsg(self, irc, msg):
                self.registryValue('allowUnaddressedKarma',
                                   msg.channel, irc.network):
                 irc = callbacks.SimpleProxy(irc, msg)
-                thing = msg.args[1].rstrip()
+                inc = self.registryValue('incrementChars', msg.channel, irc.network)
+                dec = self.registryValue('decrementChars', msg.channel, irc.network)
+                if msg.args[1].endswith(inc) or msg.args[1].endswith(dec):
+                    thing = msg.args[1].rstrip(":, ")
```
jlu5, line 321: I don't think this will actually work - if the message ends with `++` or `--` as in the default configuration, there is no way it'll also end with one of `:, `, so this is effectively a no-op.

I think this code should belong in either `_doKarma` or `_normalizeThing` so the behaviour is consistent with addressed Karma calls.

jlu5 (changes requested):

mogad0n:
bad logic

PR #1499, opened 2022-01-29
Use asyncio to manage the driver loop

progval:
This allows us to remove the custom shoehorned select() handling
without changing the driver design too much.


progval:
lol, asyncio doesn't support SSLSockets (as created by `ssl.SSLContext.wrap_socket`) so I'd need to rewrite half the Socket module to support this

PR #1504, opened 2022-03-28
Replace twitter.com with nitter.net

sillyh4ck3r:


sillyh4ck3r:
Cool.  I'll try that, instead.

PR #1508, opened 2022-06-17
Add ability to exclude channel from self censoring

SamStrongTalks:
Adds a `selfCensor` per channel setting to allow excepting a channel from the bot's censoring of its own messages.

### DIFF
--- a/plugins/BadWords/config.py
+++ b/plugins/BadWords/config.py
@@ -114,8 +114,8 @@ conf.registerGlobalValue(BadWords, 'stripFormatting',
     plugins that do coloring or bolding of text.""")))
 
 conf.registerChannelValue(BadWords, 'selfCensor',
-    registry.Boolean(True, _("""Determines whether the bot will filter itself in
-    this channel.""")))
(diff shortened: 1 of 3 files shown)


plugins/BadWords/config.py
```diff
@@ -113,6 +113,9 @@ class ReplacementMethods(registry.OnlySomeStrings):
     filtering.  If it's True, however, it will interact poorly with other
     plugins that do coloring or bolding of text.""")))
 
+conf.registerChannelValue(BadWords, 'selfCensor',
+    registry.Boolean(True, _("""Determines whether the bot will filter itself in
+    this channel.""")))
```
progval, line 118: Just "Determines whether the bot will filter its own message". Variable help strings do not mention "in this channel" because this is indicated by other means (list prefixes, current values in output of `@help list`, etc.)

progval (changes requested):

plugins/BadWords/config.py
SamStrongTalks: Done

progval:
thanks!

Issue #1490, opened 2021-09-10
Improve MoobotFactoids documentation


jlu5:
Today in #limnoria-bots we found that MoobotFactoids supports replacing and regex-replacing existing factoids. This isn't documented in the plugin README, so it's not obvious at all for folks new to the plugin.
Syntax seems to be:

- `no x is y` - replace factoid `x` with content `y`
- `x ~= s/y/z/` applies regexp `s/y/z/` to the factoid `x`

I assume that this is covered in the original moobot docs, though I'm not really sure which moobot the plugin refers to - Google gives me 3-4 different results :man_shrugging: . [blootbot](https://sourceforge.net/projects/blootbot/) (from 2005!) does have a few more examples in the tarballs, `/blootbot/doc/old/EXAMPLES`

hashborgir:
There is also : 
```
elif ['is', 'also'] in utils.seq.window(tokens, 2):
            self.augmentFactoid(irc, msg, tokens)
```

Proposed documentation change:

```
no <factoid key> is <something>
Replaced the stored factoid with a new factoid

<factoid key> ~= s/something/nothing/
Applies the sed style regex replacement to the factoid text.

<factoid key> is also <something else>
Augment the factoid, or append new text to the end of factoid.

```
Hash: secbot Hash is awesome
Hash: secbot no Hash is super awesome
Hash: secbot Hash is also likes python.
Hash: secbot Hash =~ s/awesome/amazing/

.hash
secbot: Hash is awesome
.hash
secbot: Hash is super awesome
.hash
secbot: Hash is super awesome or likes python
.hash
secbot: Hash is super amazing or likes python

Feel free to phrase it better.

PR #1511, opened 2022-07-13
for #1490 extend usage examples for moobotfactoids

mogad0n:
Attempt at #1490.

I have added usage examples which weren't documented, whilst attempting to follow the existing format and language.

I wasn't able to get a response from:
`<factoid key> ~= s/current value/new value/`
and hence I've left it undocumented.

Feel free to edit the language if desired.


progval:
thanks!

PR #1519, opened 2022-10-06
String: allow multi-character inputs in "ord" and "unicodename"

jlu5:
This allows them to support emoji ZWJ sequences, which render like one character but are actually multiple.


progval:
nice!

PR #1520, opened 2022-10-15
adding option to filter privmsg from unregistered users per network

orkim:
While running a relay bot when transitioning IRC networks I have found that I want little interaction with the bot on the old network. I can limit the channel interaction with the normal `supybot.reply.whenAddressedBy` settings. However, users can still interact with the bot via PRIVMSG.

This PR introduces a new network config option to allow the bot to ignore unregistered users (with logging of the action).  It's set to 'True' by default so no changes to users will be seen. If desired, a user can globally, or per-network, disable (set to False) the new `supybot.reply.unregistered` configuration value to ignore all interaction via PRIVMSG from unregistered users.

### DIFF
--- a/src/callbacks.py
+++ b/src/callbacks.py
@@ -111,6 +111,14 @@ def _addressed(irc, msg, prefixChars=None, nicks=None,
     nicks.insert(0, ircutils.toLower(nick))
     # Ok, let's see if it's a private message.
     if ircutils.nickEqual(target, nick):
+        if get(conf.supybot.reply.unregistered) is False:
+            try:
+                user = ircdb.users.getUser(msg.prefix)
+            except KeyError:
+                log.info("ignoring %s from unregistered user %s (%s) on %s"
+                         " network: '%s'", msg.command, msg.nick, msg.prefix,
+                         network, msg.args[1])
+                return ''
         payload = stripPrefixStrings(payload)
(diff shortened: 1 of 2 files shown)


progval:
Thanks, but this feels too specific to be in core.

What about a plugin? You should be to do it like this:

```
class MyPlugin(callbacks.Plugin):
    def inFilter(self, irc, msg):
        ignoreUnidentifiedInPrivate = self.registryValue("ignoreUnidentifiedInPrivate", network=irc.network)
        if msg.channel is None and ignoreUnidentifiedInPrivate:
            try:
                user = ircdb.users.getUser(msg.prefix)
            except KeyError:
                log.info("ignoring %s from unregistered user %s (%s) on %s"
                         " network: '%s'", msg.command, msg.nick, msg.prefix,
                         network, msg.args[1])
                msg.tag("addressed", False)
        return msg
```

orkim:
Ah, I guess I didn't look far enough into the msg.tag stuff.

Do you want me to rework this as a plugin for inclusion, or do you still feel its too specific as a plugin?

progval:
you can create your own plugin repo

or add it to https://github.com/progval/Supybot-plugins/ ; there's already IgnoreNonVoice there, and it's somewhat similar

orkim:
Thanks. Closing PR.

PR #1524, opened 2022-12-21
Fix STS parsing and handling of unchecked-TLS connections

progval:
thx to @SadieCat for reporting the issue

### DIFF
--- a/test/test_irclib.py
+++ b/test/test_irclib.py
@@ -840,79 +840,85 @@ class StsTestCase(SupyTestCase):
     def tearDown(self):
         ircdb.networks.networks = {}
 
-    def testStsInSecureConnection(self):
+    def _testStsInSecureConnection(self, cap_value):
         self.irc.driver.anyCertValidationEnabled.return_value = True
         self.irc.driver.ssl = True
         self.irc.driver.currentServer = drivers.Server('irc.test', 6697, None, False)
         self.irc.feedMsg(ircmsgs.IrcMsg(command='CAP',
-            args=('*', 'LS', 'sts=duration=42,port=12345')))
+            args=('*', 'LS', 'sts=' + cap_value)))
 
         self.assertEqual(ircdb.networks.getNetwork('test').stsPolicies, {
-            'irc.test': (6697, 'duration=42,port=12345')})
+            'irc.test': (6697, cap_value)})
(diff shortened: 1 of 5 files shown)


test/test_irclib.py
```diff
@@ -840,79 +840,85 @@ def setUp(self):
...
-            args=('*', 'LS', 'sts=duration=42,port=12345')))
+            args=('*', 'LS', 'sts=' + cap_value)))
 
         self.assertEqual(ircdb.networks.getNetwork('test').stsPolicies, {
-            'irc.test': (6697, 'duration=42,port=12345')})
+            'irc.test': (6697, cap_value)})
         self.irc.driver.reconnect.assert_not_called()
 
-    def testStsInSecureConnectionNoPort(self):
+    def testStsInSecureConnectionWithoutPort(self):
```
jlu5, line 854:
```suggestion
    def testStsInSecureConnectionWithPort(self):
```

test/test_irclib.py
```diff
@@ -840,79 +840,85 @@ def setUp(self):
...
             wait=True)
 
-    def testStsInCleartextConnection(self):
-        self.irc.driver.anyCertValidationEnabled.return_value = False
-        self.irc.driver.ssl = True
-        self.irc.driver.currentServer = drivers.Server('irc.test', 6667, None, False)
-        self.irc.feedMsg(ircmsgs.IrcMsg(command='CAP',
-            args=('*', 'LS', 'sts=duration=42,port=6697')))
+    def testStsInInsecureTlsConnectionWithPort(self):
+        self._testStsInInsecureTlsConnection('duration=42,port=6697')
```
jlu5, line 885: Out of curiosity, if we're currently on an insecure TLS connection on port 6697, and get e.g. port 12345, should we reconnect on port 12345?

jlu5 (commented):
Do we want tests for the `duration=0`, duration update and (ignored) CAP DEL cases? (Up to you)

test/test_irclib.py
jlu5: (The inline diff doesn't show it well but there were two functions named `testStsInSecureConnectionWithoutPort`)

test/test_irclib.py
progval: Unspecced, but I'd say yes.

PR #1528, opened 2023-01-12
Update README.rst

PeGaSuS-Coder:
Specify that `supybot.plugins.Services.nicks` is space delimited.


progval:
The blessed phrasing would be "Space-separated set of nicks the bot will use with services.". And the README is generated from config.py; you need to change it there first, then run `./sandbox/update_readme.sh` (or edit it directly to match)

progval:
Done in 02a0204f80f788b225b18ac4efd40c0e34bea79d and b42596a0214cddfe8ff73daa8ac00cf64e52bb81

Issue #556, opened 2014-02-22
Replies for any commands triggered through @network command go to the wrong network.


jlu5:
Let's say, for example, User1 and his Supybot is on both network1 and network2.

If User1 were to call `@network command network2 echo test` on network1, the reply of "test" would be misrouted to network2. So, if User1 was on network2 at the time (using the same nick), he would get the reply there instead of on network1. Otherwise, all replies will simply get dropped.

PR #1540, opened 2023-06-04
Route commands from Network.command back to the original network

jlu5:
Add a replyIrc parameter to ReplyIrcProxy to run a command on one network, but route the replies to another.

This fixes a long standing issue GH-556 where replies for remote commands are often lost to the void, as the nick of the caller may not exist on the target network (or worse, it could belong to a completely unrelated person).


progval (commented):
Amazing it's that simple, thanks!

progval (approved):

PR #1544, opened 2023-07-01
Deprecate using supybot.dynamicScope without an import

jlu5:
Some notes:
- I'm using UserWarning here because Python doesn't show DeprecationWarnings to the console by default. (I'm not sure if changing that would be a better option)


jlu5:
I will probably defer this until we drop Python < 3.8 (Python 3.7 went [EOL in June 2023](https://endoflife.date/python)), as those versions don't support the stacklevel argument to `log.warning`. I don't think it's worth adding compat code for these old branches given the current timeframe

progval:
Could you update core and built-in plugins to use an import, so they don't display these warnings?

PR #1546, opened 2023-07-01
Add support for account-based channel bans

progval:
using https://github.com/ircv3/ircv3-specifications/pull/464


src/conf.py
```diff
@@ -1215,33 +1215,87 @@ def makeBanmask(self, hostmask, options=None, channel=None, network=None):
...
+                network, channel)()
+        options = [option for option in options if option != 'account']
+        print(hostmask, options)
+        masks = self.makeExtBanmasks(
+            hostmask, options, channel, network=network)
+        assert len(masks) == 1, 'Unexpected number of banmasks: %r' % masks
+        print(masks)
+        return masks[0]
+
+    def makeExtBanmasks(self, hostmask, options=None, channel=None, *, network):
```
jlu5, line 1238: `s/options/banstyles/` perhaps?

src/conf.py
```diff
@@ -1215,33 +1215,87 @@ def makeBanmask(self, hostmask, options=None, channel=None, network=None):
...
         bhost = '*'
         if not options:
             options = supybot.protocols.irc.banmask.getSpecific(
                 network, channel)()
+
+        add_star_mask = False
+        masks = []
+
         for option in options:
             if option == 'nick':
```
jlu5, line 1269: I usually prefer Enum types instead of fixed string inputs, especially if we silently ignore invalid options.

plugins/Channel/test.py
```diff
@@ -161,9 +163,13 @@ def testVoice(self):
...
                         m.args == (self.channel, '+v', 'bar'))
 
-    def assertKban(self, query, hostmask, **kwargs):
+    def assertKban(self, query, *hostmasks, **kwargs):
         m = self.getMsg(query, **kwargs)
-        self.assertEqual(m, ircmsgs.ban(self.channel, hostmask))
+        self.assertIn(m, [
+            ircmsgs.bans(self.channel, permutation)
+            for permutation in itertools.permutations(hostmasks)
+            ])
```
jlu5, line 171: Would https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual work better here?

src/irclib.py
```diff
@@ -680,6 +680,12 @@ class IrcState(IrcCommandDispatcher, log.Firewalled):
 
         Stores the last hostmask of a seen nick.
 
+        :type: ircutils.IrcDict[str, str]
+
+    .. attribute:: nicksToAccounts
+
+        Stores the current network account name of a seen nick.
```
jlu5, line 687:
```suggestion
        Stores the current services account name of a seen nick.
```

plugins/Channel/test.py
```diff
@@ -219,11 +225,88 @@ def join():
...
+        def join():
+            self.irc.feedMsg(ircmsgs.join(
+                self.channel, prefix='foobar!user@host.domain.tld'))
+        join()
+        self.irc.feedMsg(ircmsgs.op(self.channel, self.irc.nick))
+        self.assertKban('kban --account --exact foobar',
+                        'foobar!user@host.domain.tld')
+        join()
+        self.assertKban('kban --account foobar',
+                        'foobar!user@host.domain.tld')
```
jlu5, line 242: I wonder if it'd be better to throw an error if only `--account` is given and the user's not logged in. Exact match bans are pretty trivial to evade on most networks (even accidentally if your connection times out), to the point I doubt there's much value in the bot setting them in the first place.

src/irclib.py
```diff
@@ -957,6 +979,11 @@ def do354(self, irc, msg):
         (n, t, user, ip, host, nick, status, account, gecos) = msg.args
         hostmask = '%s!%s@%s' % (nick, user, host)
         self.nicksToHostmasks[nick] = hostmask
+        if account == '0':
+            # logged out
+            self.nicksToAccounts[nick] = None
```
jlu5, line 984: Storing None values seems a bit odd, would it be better to delete the entry?

src/ircutils.py
```diff
@@ -345,6 +345,25 @@ def banmask(hostmask):
         else:
             return '*!*@'  + host
 
+
+def accountExtban(nick, irc):
```
jlu5, line 349: nit: having `irc` as the first argument is more consistent with the other functions here

test/test_irclib.py
```diff
@@ -579,6 +579,105 @@ def testChghost(self):
...
         self.assertEqual(st.nickToHostmask('foo'), 'foo!bar2@baz2')
 
+    def testNickToAccountBaseJoin(self):
+        st = irclib.IrcState()
+
+        st.addMsg(self.irc, ircmsgs.join('#foo', prefix='foo!bar@baz'))
+        st.addMsg(self.irc, ircmsgs.join('#foo', prefix='bar!baz@qux'))
+        with self.assertRaises(KeyError):
+            st.nickToAccount('foo')
+            st.nickToAccount('bar')
```
jlu5, line 589: Are we expecting both calls to raise, or one of the two?

test/test_irclib.py
```diff
@@ -579,6 +579,105 @@ def testChghost(self):
...
+            st.nickToAccount('qux')
+
+        # QUIT erases the entry
+        with self.subTest("QUIT"):
+            st2 = st.copy()
+            st2.addMsg(self.irc, ircmsgs.quit(prefix='foo!bar@baz'))
+            with self.assertRaises(KeyError):
+                st2.nickToAccount('foo')
+            self.assertEqual(st2.nickToAccount('bar'), 'account2')
+            self.assertEqual(st.nickToAccount('foo'), 'account1')
```
jlu5, line 618: Check `st2` instead of `st`?

test/test_irclib.py
```diff
@@ -579,6 +579,105 @@ def testChghost(self):
...
+        # NICK moves the entry
+        with self.subTest("NICK"):
+            st2 = st.copy()
+            st2.addMsg(self.irc, ircmsgs.IrcMsg(prefix='foo!bar@baz',
+                                                command='NICK', args=['foo2']))
+            with self.assertRaises(KeyError):
+                st2.nickToAccount('foo')
+            self.assertEqual(st2.nickToAccount('foo2'), 'account1')
+            self.assertEqual(st2.nickToAccount('bar'), 'account2')
+            self.assertEqual(st.nickToAccount('foo'), 'account1')
```
jlu5, line 630: Same as above

test/test_irclib.py
```diff
@@ -579,6 +579,105 @@ def testChghost(self):
...
+
+        # NICK moves the entry (and overwrites if needed)
+        with self.subTest("NICK with overwrite"):
+            st2 = st.copy()
+            st2.addMsg(self.irc, ircmsgs.IrcMsg(prefix='foo!bar@baz',
+                                                command='NICK', args=['bar']))
+            with self.assertRaises(KeyError):
+                st2.nickToAccount('foo')
+            self.assertEqual(st2.nickToAccount('bar'), 'account1')
+            self.assertEqual(st.nickToAccount('foo'), 'account1')
```
jlu5, line 641: Same as above

src/conf.py
```diff
@@ -1215,33 +1215,87 @@ def makeBanmask(self, hostmask, options=None, channel=None, network=None):
...
+
+        if add_star_mask and (bnick, buser, bhost) != ('*', '*', '*'):
+            masks.append(ircutils.joinHostmask(bnick, buser, bhost))
+
         if (bnick, buser, bhost) == ('*', '*', '*') and \
-                ircutils.isUserHostmask(hostmask):
-            return hostmask
-        return ircutils.joinHostmask(bnick, buser, bhost)
+                ircutils.isUserHostmask(hostmask) and \
+                masks == []:
```
jlu5, line 1294:
```suggestion
                not masks:
```

src/irclib.py
progval: no entry means "unknown", `None` entry means "logged out"

jlu5: It might be nice to have the function describe how it'll throw KeyError if the user isn't in the state

test/test_irclib.py
progval: nope, this is meant to check `st` isn't affected by a message added to `st2`

progval: I added a comment to clear the confusion

src/conf.py
progval: it's meant to be existent with the existing `makeBanmask` :/

src/conf.py
progval: ditto

progval: I guess I can allow both and deprecate the string though.

src/ircutils.py
progval: Right. I probably wanted to make it optional or something, but I don't think it makes sense.

src/conf.py
jlu5: Ah I missed that, not a big deal then

plugins/Channel/test.py
progval: Heh, I don't know

```diff
diff --git a/plugins/Channel/test.py b/plugins/Channel/test.py
index bd716767c..8b08df6e3 100644
--- a/plugins/Channel/test.py
+++ b/plugins/Channel/test.py
@@ -165,10 +165,9 @@ def testVoice(self):
 
     def assertKban(self, query, *hostmasks, **kwargs):
         m = self.getMsg(query, **kwargs)
-        self.assertIn(m, [
-            ircmsgs.bans(self.channel, permutation)
-            for permutation in itertools.permutations(hostmasks)
-            ])
+        self.assertEqual(m.command, "MODE", m)
+        self.assertEqual(m.args[0], "+" + "b" * len(hostmasks), m)
+        self.assertCountEqual(m.args[1], hostmasks, m)
 
         m = self.getMsg(' ')
         self.assertEqual(m.command, 'KICK')
```

src/conf.py
progval: Heh, I'll just log invalid options instead.

plugins/Channel/test.py
progval: What if it uses the default banmask instead?

progval: I made it fall back to the default banmask, then to just the host (in case the default banmask is only `['account']` for some reason)

progval:
Should be all good now. The spec was just merged by IRCv3 (as a draft) so I'm going to merge this PR soon-ish

PR #1552, opened 2023-09-03
Add command to manually initiate SASL

glguy:


plugins/Admin/plugin.py
```diff
@@ -359,6 +359,20 @@ def acmd(self, irc, msg, args, commandAndArgs):
...
+        @internationalizeDocstring
+        def start(self, irc, msg, args):
+            """takes no arguments
+
+            Manually initiates SASL authentication
+            """
+            if 'sasl' in irc.state.capabilities_ack:
+                irc.startSasl(msg)
+            else:
+                irc.requestCapabilities(['sasl'])
```
progval, line 373: Let's just add it to `REQUEST_CAPABILITIES` unconditionally, so you can return an error here if it's not negotiated.

plugins/Admin/plugin.py
```diff
@@ -359,6 +359,20 @@ def acmd(self, irc, msg, args, commandAndArgs):
             self.Proxy(irc.getRealIrc(), msg, commandAndArgs)
     acmd = wrap(acmd, ['admin', many('something')])
 
+    class sasl(callbacks.Commands):
+
+        @internationalizeDocstring
+        def start(self, irc, msg, args):
```
progval, line 365: Move it to the Network plugin

src/irclib.py
```diff
@@ -1767,6 +1757,19 @@ def resetSasl(self):
...
+        self.sasl_ecdsa_key = network_config.sasl.ecdsa_key()
+        self.sasl_scram_state = {'step': 'uninitialized'}
+        self.authenticate_decoder = None
+        self.sasl_next_mechanisms = []
+        self.sasl_current_mechanism = None
+
+        self.resetSaslMechanisms(network_config)
+
         if self.sasl_next_mechanisms:
             self.REQUEST_CAPABILITIES.add('sasl')
```
progval, line 1774: and you can remove that bit

src/irclib.py
```diff
@@ -1913,6 +1916,22 @@ def _maybeStartSasl(self, msg):
                         if x.lower() in available]
             self.tryNextSaslMechanism(msg)
 
+    def startSasl(self, msg):
```
progval, line 1919: You should be able to dedup this function with `_maybeStartSasl` above

progval (changes requested):
Thanks!

plugins/Admin/plugin.py
jlu5: I would recommend renaming this to something like `sasl auth` or `sasl login`. `start` sounds like something that starts / enables a long running task, which isn't really the case here. (There's no corresponding `stop` command for it, for instance)

jlu5 (changes requested):

plugins/Admin/plugin.py
jlu5: I would consider renaming this to something like `sasl auth` or `sasl login`. `start` is less descriptive and sounds like something that starts a longer running task or service, which isn't the case here.

src/irclib.py
glguy: I *think* we should unconditionally call resetSasl() when starting a new SASL transaction as I've done here. But I'm not quite certain as this function kind of mixes configuration and the sasl state machine state initialization. Is this an issue? Do I need to split those in half?

plugins/Network/plugin.py
```diff
@@ -306,6 +306,17 @@ def capabilities(self, irc, msg, args, otherIrc):
...
+        """takes no arguments
+
+        Manually initiate SASL authentication.
+        """
+        if 'sasl' in irc.state.capabilities_ack:
+            irc.startSasl(msg)
+            irc.replySuccess()
+        else:
+            irc.error(_('SASL not supported'))
+    saslauth = wrap(saslauth)
```
progval, line 319:
```suggestion
    def authenticate(self, irc, msg, args):
        """takes no arguments

        Manually initiate SASL authentication.
        """
        if 'sasl' in irc.state.capabilities_ack:
            irc.startSasl(msg)
            irc.replySuccess()
        else:
            irc.error(_('SASL not supported'))
    authenticate = wrap(authenticate)
```

we avoid concatenating words in command names; so it has to be either a single a single word or space-separated. I don't have a strong opinion on which is best here.

progval (changes requested):
One last thing

src/irclib.py
progval: seems fine

progval:
Could you rewrite your history to have the test changes first, then squash the rest together? This will make history easier to read and bisect

progval:
```diff




























diff --git a/test/test_irclib.py b/test/test_irclib.py
index 18f60b53d..2f79b2f34 100644
--- a/test/test_irclib.py
+++ b/test/test_irclib.py
@@ -1494,12 +1494,16 @@ def testBatchInterleaved(self):
 
 class SaslTestCase(SupyTestCase, CapNegMixin):
     def setUp(self):
-        pass
+        self._default_mechanisms = conf.supybot.networks.test.sasl.mechanisms()
+
+    def tearDown(self):
+        conf.supybot.networks.test.sasl.mechanisms.setValue(self._default_mechanisms)
 
     def testPlain(self):
         try:
             conf.supybot.networks.test.sasl.username.setValue('jilles')
             conf.supybot.networks.test.sasl.password.setValue('sesame')
+            conf.supybot.networks.test.sasl.mechanisms.setValue(['external', 'plain'])
             self.irc = irclib.Irc('test')
 
             self.startCapNegociation()
@@ -1526,11 +1530,11 @@ def testExternalFallbackToPlain(self):
         try:
             conf.supybot.networks.test.sasl.username.setValue('jilles')
             conf.supybot.networks.test.sasl.password.setValue('sesame')
+            conf.supybot.networks.test.sasl.mechanisms.setValue(['external', 'plain'])
             conf.supybot.networks.test.certfile.setValue('foo')
             self.irc = irclib.Irc('test')
 
             self.assertEqual(self.irc.sasl_current_mechanism, None)
-            self.irc.sasl_next_mechanisms = ['external', 'plain']
 
             self.startCapNegociation()
```

works for me after this patch

PR #1556, opened 2023-09-18
Implement changing of sed response per channel (fixes #1433)

### LINKED ISSUE #1433: Configurable string for sedregex-plugin responses per channel
it would be nice to be able to alter the 'static introduction' per chan from sedregex-plugin, since i want to conf different phrases/langs instead of 'meant to say' per #chan

famfo:


progval:
Thanks! Could you also make the `'%s thinks %s'` part configurable?

In order for it to make sense, you can use two independent config variables, ie. `$nick meant to say: $replacement` and `$otherNnick thinks $nick meant to say: $replacement` and `ircutils.standardSubstitute(irc, m, self.registryValue(...), {'otherNick': msg.nick, 'replacement': subst}`) which gives more flexibility to bot admins (they can use `$host`, date, etc.)

famfo:
Sure, I can look into it.

plugins/SedRegex/plugin.py
```diff
@@ -239,8 +235,15 @@ def _replacer_process(self, irc, msg, target, pattern, replacement, count, messa
 
                         subst = axe_spaces(subst)
 
-                        return _("%s meant to say: %s") % \
-                            (messageprefix, subst)
+                        if m.nick == msg.nick:
+                            sed_string = self.registryValue('sedString', msg.channel, irc.network)
+                            return _("%s%s %s") % \
+                                (m.nick, sed_string, subst)
```
jlu5, line 241: Please use [template strings](https://docs.python.org/3/library/string.html#template-strings) as val suggested. If you do this with [`ircutils.standardSubstitute`](https://github.com/progval/Limnoria/blob/5357f50bed9a830994faf663416c4b05b21f00b0/src/ircutils.py#L1024) (pass the relevant variables here as `env`) it not only allows you to use more variables that the bot fills in at runtime, but also lets you change the entire output string in a freeform way.

jlu5 (changes requested):
Can you add tests for the additional variables to SedRegex/test.py ?

plugins/SedRegex/plugin.py
famfo: I don't fully understand how I should do this (especially why I do not entirely understand the standardSubstitute)

jlu5: Basically you call it as `ircutils.standardSubstitute(irc, msg, templateString, env)`, where templateString is something like `$caller thinks $targetNick meant to say: $subst`, and `env` is a dictionary of template variables to values (e.g. `{'targetNick': m.nick, 'caller': msg.nick, 'replacement': subst}`). You can make the config variable hold the entire template string, and this means you can configure the output however you want. A less verbose `<$targetNick> $subst` is possible too for example

famfo: Implemented in my latest commit.

plugins/SedRegex/plugin.py
```diff
@@ -239,8 +235,15 @@ def _replacer_process(self, irc, msg, target, pattern, replacement, count, messa
...
-                        return _("%s meant to say: %s") % \
-                            (messageprefix, subst)
+                        if m.nick == msg.nick:
+                            fmt = self.registryValue('sedString', msg.channel, irc.network)
+                            env = {'nick': m.nick, 'replacement': subst}
+                        else:
+                            fmt = self.registryValue('sedEdited', msg.channel, irc.network)
+                            env = {'nick': m.nick, 'otherNick': msg.nick, 'replacement': subst}
+
+                        return ircutils.standardSubstitute(irc, msg, fmt, env)
```
progval, line 245:
```suggestion
                        if m.nick == msg.nick:
                            fmt = self.registryValue('sedString', msg.channel, irc.network)
                            env = {'replacement': subst}
                        else:
                            fmt = self.registryValue('sedEdited', msg.channel, irc.network)
                            env = {'otherNick': msg.nick, 'replacement': subst}

                        return ircutils.standardSubstitute(irc, m, fmt, env)
```

`standardSubstitute` will automatically set `$nick` from the `m` message

famfo: Oh okay, didn't know that, I'll change that.

plugins/SedRegex/config.py
```diff
@@ -57,6 +57,17 @@ def configure(advanced):
 conf.registerChannelValue(SedRegex, 'ignoreRegex',
     registry.Boolean(True, _("""Should Perl/sed regex replacing
                      ignore messages which look like valid regex?""")))
+conf.registerChannelValue(SedRegex, 'sedString',
```
jlu5, line 60: I would name this variable `formatString` and the other one `formatString.other` so that their meaning is clearer just from their names. (To register a variable under another one you could do something like `conf.registerChannelValue(SedRegex.formatString, ...)`

LGTM otherwise!

jlu5 (changes requested):

plugins/SedRegex/config.py
progval: Just `format` instead of `formatString`

plugins/SedRegex/config.py
```diff
@@ -57,6 +57,17 @@ def configure(advanced):
 conf.registerChannelValue(SedRegex, 'ignoreRegex',
     registry.Boolean(True, _("""Should Perl/sed regex replacing
                      ignore messages which look like valid regex?""")))
+conf.registerChannelValue(SedRegex, 'format',
+    registry.String(_('$nick meant to say: $replacement'), _("""Sets the format
+                                    string for a message edited by the original
+                                    author. Required fileds: $nick (nick of the
```
jlu5, line 63:
```suggestion
                                    author. Required fields: $nick (nick of the
```

plugins/SedRegex/config.py
```diff
@@ -57,6 +57,17 @@ def configure(advanced):
...
+                                    author), $replacement (edited message)""")))
+conf.registerChannelValue(SedRegex.format, 'other',
+    registry.String(_('$otherNick thinks $nick meant to say: $replacement'), _("""
+                                    Sets the format string for a message edited by
+                                    another author. Required fileds: $nick (nick
```
jlu5, line 68:
```suggestion
                                    another author. Required fields: $nick (nick
```

plugins/SedRegex/test.py
```diff
@@ -279,6 +279,23 @@ def testSlashInNicks(self):
         with conf.supybot.protocols.irc.strictRfc.context(True):
             self.assertSnarfNoResponse('%s: s/123/321/' % ircutils.nickFromHostmask(frm), frm=self.__class__.other2)
 
+    def testFmtString(self):
+        fmt = "<$nick>: $replacement"
+        with conf.supybot.plugins.sedregex.format.context(fmt):
+            self.feedMsg('frog')
+            self.feedMsg('s/frog/frogged')
```
jlu5, line 286:
```suggestion
            self.feedMsg('s/frog/frogged/')
```

jlu5: (I consider missing trailing slash a separate case to test)

jlu5 (changes requested):

PR #1558, opened 2023-09-26
Feature/daily scheduler events

T-101:
Extend Scheduler plugin to run commands daily

### DIFF
--- a/plugins/Scheduler/plugin.py
+++ b/plugins/Scheduler/plugin.py
@@ -33,6 +33,7 @@ import os
 import math
 import shutil
 import tempfile
+import datetime
 
 import supybot.conf as conf
 import supybot.utils as utils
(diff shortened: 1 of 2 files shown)


progval:
`@daily name HH:MM:SS <command>` is equivalent to `@repeat --delay [seconds HH:MM:SS] name [seconds 1d] <command>`, and shares its main issue: the event runs every 86400 seconds instead of every day, which does not behave as expected in presence of DST and leap seconds.

In order to address this, you need to add a new type of scheduler events for events which happens on a fixed clock (every X days, monthly, yearly, ...) and compute the next tick every time.

(Also, avoid loops to compute time, it uses more CPU and is more error prone, eg. this one does not terminate when a user provides `12:00:60` as time_of_day)

PR #1560, opened 2023-09-29
requirements.txt: add ddate as an optional dependency for Time.ddate

Mikaela:
The command was introduced in https://github.com/progval/Limnoria/pull/1310 and I imagine it missing from `requirements.txt` being an oversight.


progval:
thx

Issue #1455, opened 2021-05-24
supybot.reply.whenAddressedBy.nick.atEnd: there is no reply if there is space after the nick


Mikaela:
I like having the bot reply also when its nick is at the end of message and have configured mine to only reply to nick on `#Limnoria-bots`, where I noticed that there is no response when there is a space in the end which was confusing me.

I think there should be a reply regardless of the space as I think IRC clients generally add spaces when tab-completing nicknames.

```
2021-144 15:11:48 EEST <Mikaela> version R-66Y 
2021-144 15:11:52 EEST <Mikaela> version R-66Y
2021-144 15:11:53 EEST <R-66Y> The current (running) version of this Limnoria is 2021.05.22, running on Python 3.7.3 (default, Jan 22 2021, 20:04:44)  [GCC 8.3.0].  The newest versions available online are 2021.05.22 (in master), 2021.05.22 (in testing).
```

PR #1563, opened 2023-10-05
Ignore trailing whitespace when addressing the bot by nick at end

matiasw:
Seems to fix https://github.com/progval/Limnoria/issues/1455 

Again, if you decide to merge, please add the `hacktoberfest-accepted` label.

### DIFF
--- a/test/test_callbacks.py
+++ b/test/test_callbacks.py
@@ -269,7 +269,6 @@ class FunctionsTestCase(SupyTestCase):
         irc = getTestIrc()
         msg = ircmsgs.privmsg('#foo', 'baz, bar')
         irc._tagMsg(msg)
-        nick = irc.nick
         self.assertEqual(callbacks.addressed('bar', msg,
                                              whenAddressedByNickAtEnd=True),
                          'baz')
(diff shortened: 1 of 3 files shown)


progval:
Could you add a test for this?

matiasw:
Sure; I added it under the Misc plugin, as that's where this was originally reported.

progval:
It should be in `tests/test_callbacks.py`, that's where the other tests for this are

matiasw:
> It should be in `tests/test_callbacks.py`, that's where the other tests for this are

How do I run these tests? `supybot-test` seems to be only for plugins, and there is no `test` target in the Makefile.

progval:
`supybot-test test`

> there is no test target in the Makefile.

There is, and it runs both the above and plugin tests

matiasw:
Ok; I moved & updated the test.

test/test_callbacks.py
```diff
@@ -269,9 +269,16 @@ def testAddressedWithNickAtEnd(self):
         irc = getTestIrc()
         msg = ircmsgs.privmsg('#foo', 'baz, bar')
         irc._tagMsg(msg)
+        nick = irc.nick
```
progval, line 272:
```suggestion
```

matiasw: Sorry! ;(

progval:
thx

Issue #1061, opened 2015-02-25
supybot-test dumps temporary folders into the working directory


NP-Hardass:
The current behavior when running supybot-test is to create several folders and files in the current working directory.  This has several side effects.  supybot-test cannot be run while the user doesn't have write permissions to cwd, supybot cannot be run more than once from within plugins-dir because it attempts to interpret its own leftover test directories as plugins, and supybot-test has the potential to clobber other files/folders when running in a directory in which the folders that it uses already exist. 
The temporary files/folders created by the supybot-test are left behind.  If there is a legitimate use for them, then they should be removed in all instances except where a flag is raised (so that normal users aren't burdened with extra files/folders) 

``` pseudocode
if user can write to /tmp
  TESTDIR=`mktemp -d`
else
  mkdir -p ~/.tmp
  TESTDIR=`mktemp -d -p  ~/.tmp`
create temp folders/files in TESTDIR
run test suite
if keep testdir flag
  rm -rf TESTDIR
```

Mikaela:
Concerns:
* Cross-platform support
    * Windows
* being Python

NP-Hardass:
Sure...
I gave an example of good behavior, not a final implementation...

ProgVal:
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

jamessan:
The main reason the files aren't deleted is so that there's information to inspect when there are test failures.  Granted, the default behavior could be to remove the files if everything passes.

PR #1570, opened 2024-04-12
Files are created in tmp dirs and they are cleaned if tests succeed (#1061)

GMDSantana:
As discussed in #1061, here we use this:

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

to handle the creation of temporary directories. The files are not deleted in case there are any test failures, so they can be inspected.

### DIFF
--- a/src/scripts/limnoria_test.py
+++ b/src/scripts/limnoria_test.py
@@ -45,7 +45,6 @@ import traceback
 
 # We need to do this before we import conf.
 main_temp_dir = TemporaryDirectory()
-print(f"Temporary directory path: {main_temp_dir.name}")
 
 os.makedirs(os.path.join(main_temp_dir.name, 'conf'))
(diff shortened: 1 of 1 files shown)


progval:
Thanks

Could you create a single temporary directory with the three `test-` inside it, as well as print its path? This will make it easier for people running tests to locate where the data is.

src/scripts/limnoria_test.py
```diff
@@ -36,23 +36,28 @@
...
 
 import supybot
 import logging
 import traceback
 
 # We need to do this before we import conf.
-if not os.path.exists('test-conf'):
-    os.mkdir('test-conf')
+main_temp_dir = TemporaryDirectory()
+print(f"Temporary directory path: {main_temp_dir.name}")
```
progval, line 48: Move this to the `.detach()` case so it's printed at the end (more visible), and only if the directory is kept.

progval:
Thanks!

PR #1573, opened 2024-05-25
Math: Round trailing nines up in `@convert`

progval:


jlu5:
I'm honestly surprised that this code is so complicated. Would anyone actually complain if we used a fixed (or configurable?) max precision like 2 or 3 decimal places?

progval:
I would.

2 is definitely too little (8/9 = 1.125) so I wouldn't feel comfortable with 3 either.

The complexity is because it tries to fix physics calculous rules, ie. output should have as many significant digits as the input

jlu5 (approved):

PR #1579, opened 2024-06-21
Fix: Ignore trailing chars/spaces for Karma

prdes:
The trailing characters `,` and `:` have been chosen as they are the most popular ones appended to tab completed nicks.

Solves #1495


progval:
Could you add `\t` too?

PR #1585, opened 2024-07-24
Added russian locale to limnoria

ssdaniel24:
Translated with lokalise.com. So export settings were a little limited (didn't find how to break lines to fit in 80 columns).

Translation is full, but has some *fuzzy* moments, will improve it in a future.

### DIFF
--- a/locales/ru.po
+++ b/locales/ru.po
@@ -1,0 +1,774 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"POT-Creation-Date: 2022-02-06 00:12+0100\n"
+"PO-Revision-Date: 2024-06-17 13:29+0300\n"
+"Last-Translator: ssdaniel24 <bo7oaonteg2m__at__mailDOTru>\n"
+"Language-Team: \n"
+"Language: ru\n"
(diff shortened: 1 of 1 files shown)


locales/ru.po
```diff
@@ -0,0 +1,774 @@
...
+msgstr "Это не является допустимым %s."
+
+msgid "(XX more messages)"
+msgstr "(ещё XX сообщений)"
+
+msgid "more message"
+msgstr "сообщение осталось"
+
+msgid "more messages"
+msgstr "сообщений осталось"
```
progval, line 46: these two translations need to have fewer bytes bytes as the translation of "XX more messages" minus the leading "XX " (preferably as many), because they are used in the same context. (the number is prepended to these two strings)

progval (commented):
Thanks! I don't speak Russian but I appreciate that it's a lot of work to do this :)

locales/ru.po
progval: You don't have to, you just need to replace `(ещё XX сообщений)` with `(XX сообщений осталось)`, so that both translations have the same number of bytes in the end

PR #1591, opened 2024-08-25
Updated russian translation

ssdaniel24:
Update based on updated `message.pot`.

*P.S. but it seems like nothing new? Possibly poedit just touched file with its formatting...*


progval:
yeah looks like msgmerge reformatted the file, which undid poedit's own reformatting

PR #1595, opened 2024-10-04
[Unix] fix: add space to avoid breaking ipv6 formating

codeurimpulsif:
## Goal

This PR just add a space in a split to avoid splitting string on first `:` occurence inside an IPv6 with the `ping` command.


progval:
thanks!

PR #1598, opened 2024-10-18
CI: Replace Python 3.13.0-alpha.6 with 3.13.0

progval:


progval:
Currently blocked by https://github.com/python/cpython/issues/125714#issuecomment-2423266138 because `commands.process()` relies on an unpickable target

jlu5 (approved):

PR #1599, opened 2024-10-19
Add support for Python 3.14.0-alpha1

progval:


progval:
This is tough: Python 3.14 switches `multiprocessing`'s default start method from `fork` to `forkserver`, which requires `target` functions to be picklable.

The above commits fix the low-hanging fruits, but there is still the issue of plugins using `target=self._some_method`, because methods reference their plugin class, which can only be unpickled if the module containing the plugin class can be imported. And generally it can't, because plugin modules are not on Python's import path, as they are imported with this:

https://github.com/progval/Limnoria/blob/54c09809786db7a6468c48dedc788287fbcded72/src/plugin.py

progval:
> Of course we can change the start method back to `fork` when Limnoria starts, but it doesn't solve the issue that it is [unsafe in threaded programs](https://docs.python.org/3.14/library/os.html#os.fork) (which Limnoria is)

Let's do that. I opened https://github.com/progval/Limnoria/issues/1600 about this

PR #1605, opened 2024-12-06
Fix interference of wrap() and internationalizeDocstring()

progval:
Most commands are decorated with @internationalizeDocstring then with wrap(). wrap() itself calls internationalizeDocstring(), so that plugins wrapping with @internationalizeDocstring() is now optional; but many still do it.

This means that docstrings went twice through the
_PluginInternationalization.

This fixes the bug that on the second run, it would try to translate again the translated message, which fails (because the translated message is not in English); and then fell back to the original string (which is in English).

This commit changes the behavior to return the already-translated string directly instead.

### DIFF
--- a/src/i18n.py
+++ b/src/i18n.py
@@ -272,7 +272,7 @@ class _PluginInternationalization:
         try:
             string = self._translate(normalizedUntranslated)
         except KeyError:
-            string = untranslated
+            string = originalUntranslated
         return self._addTracker(string, untranslated)
 
     def _translate(self, string):
--- a/src/i18n.py
+++ b/src/i18n.py
@@ -1,5 +1,5 @@
 ###
-# Copyright (c) 2010-2021, Valentin Lorentz
+# Copyright (c) 2010-2024, Valentin Lorentz
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


src/i18n.py
```diff
@@ -263,16 +263,17 @@ def __call__(self, untranslated):
         """Main function.
 
         This is the function which is called when a plugin runs _()"""
-        normalizedUntranslated = normalize(untranslated, True)
+        if untranslated.__class__ is InternationalizedString:
```
jlu5, line 266: Would `isinstance(untranslated, InternationalizedString)` be more idiomatic?

progval: it is, but it's noticeably slower:

```
In [1]: from supybot.i18n import InternationalizedString

In [2]: s1 = "foo bar"

In [3]: s2 = InternationalizedString("foo bar")

In [4]: %timeit s1.__class__ is InternationalizedString
20.7 ns ± 0.043 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit isinstance(s1, InternationalizedString)
39 ns ± 0.237 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit s2.__class__ is InternationalizedString
21.4 ns ± 0.0408 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: %timeit isinstance(s2, InternationalizedString)
26.7 ns ± 0.0825 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [8]: %timeit s1.__class__ is InternationalizedString
20.9 ns ± 0.0671 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: %timeit isinstance(s1, InternationalizedString)
38.9 ns ± 0.164 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [10]: %timeit s2.__class__ is InternationalizedString
21.4 ns ± 0.0499 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [11]: %timeit isinstance(s2, InternationalizedString)
26.4 ns ± 0.0881 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

I've seen non-Limnoria Python code where `isinstance` is a bottleneck; and I wouldn't be surprised if it happened to some plugin doing strange things with strings.

jlu5 (approved):

PR #1608, opened 2024-12-07
Added --random flag for 'rss' command

### LINKED ISSUE #1593: RSS: add 'random' command
Add 'random' command which will parse the feed and return random item from it.

ssdaniel24:
Added `--random` flag for 'rss' command. It is randomizing initial order of rss entries. So you get random entry of given rss feed after every new call with this flag.

Fixes #1593.

### DIFF
--- a/plugins/RSS/plugin.py
+++ b/plugins/RSS/plugin.py
@@ -36,6 +36,7 @@ import json
 import time
 import types
 import string
+import random
 import socket
 import threading
 import feedparser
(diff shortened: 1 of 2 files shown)


plugins/RSS/plugin.py
```diff
@@ -719,17 +724,22 @@ def rss(self, irc, msg, args, url, n):
...
             irc.error(s)
             return
-        n = n or self.registryValue('defaultNumberOfHeadlines', channel, irc.network)
         entries = list(filter(lambda e:self.should_send_entry(irc.network, channel, e),
                               feed.entries))
-        entries = sort_feed_items(entries, 'newestFirst')
+        upper_sort = 'newestFirst'
+        if [y for x,y in optlist if x == 'random']:
+            upper_sort = 'random'
+        entries = sort_feed_items(entries, upper_sort)
```
progval, line 732:
```suggestion
        order = 'newestFirst'
        if 'random' in (x for x,y in optlist):
            order = 'random'
        entries = sort_feed_items(entries, order)
```

looks more readable to me

progval (commented):
thanks!

PR #1609, opened 2024-12-14
Clarify semantics of vhost/vhostv6

progval:


src/conf.py
```diff
@@ -422,11 +422,17 @@ def registerNetwork(name, password='', ssl=True, sasl_username='',
     registerGlobalValue(network, 'vhost',
         registry.String('', _("""Determines what vhost the bot will bind to before
         connecting a server (IRC, HTTP, ...) via IPv4. If empty, defaults to
-        supybot.protocols.irc.vhost""")))
+        supybot.protocols.irc.vhost. It must be (or resolve to) an IP address
+        assigned to one of the network interfaces (see 'ip addr' on Linux).
+        This may differ from the hostname visible on IRC, which is based on the
```
jlu5, line 427: Instead of "outgoing IP address" (which isn't technically accurate) I would say something like:

This may differ from your public IP address, if you are behind a NAT.

progval: thx

jlu5 (approved):

PR #1613, opened 2025-02-01
Added oEmbed support to the Web plugin to improve title fetching

lodriguez:
Two methods are supported and can be configured:
1. Registry lookup via oembed.com (useOembedRegistry)
2. HTML discovery of endpoints (useOembedDiscovery)

Registry is tried first if enabled, then discovery if enabled,
finally falling back to HTML parsing.

_I think it would be nicer to add the calls to `getTitle()`._
_I was trying to avoid using `url_workaround()`, since it would break support for reddit - for exsample, old.reddit.com isn't in the registry. However, following redirects could actually be beneficial, especially for handling v.reddit.com._
_For discovery, it makes sense to place this logic close to the title parser, as that would save an extra request. That said, I wanted to open the pull request first to gather some feedback before making further adjustments._

### DIFF
--- a/plugins/Web/config.py
+++ b/plugins/Web/config.py
@@ -95,4 +95,12 @@ conf.registerGlobalValue(Web.fetch, 'timeout',
     seconds the bot will wait for the site to respond, when using the 'fetch'
     command in this plugin. If 0, will use socket.defaulttimeout"""))
 
+conf.registerGlobalValue(Web, 'useOembedRegistry',
+    registry.Boolean(False, _("""Determines whether the bot will use the 
+    oembed.com providers registry.""")))
+
+conf.registerGlobalValue(Web, 'useOembedDiscovery',
+    registry.Boolean(False, _("""Determines whether the bot will use HTML
+    discovery to find oEmbed endpoints.""")))
+
 # vim:set shiftwidth=4 softtabstop=4 expandtab textwidth=79:
--- a/plugins/Web/plugin.py
+++ b/plugins/Web/plugin.py
@@ -296,17 +296,30 @@ class Web(callbacks.PluginRegexp):
                     regex = re.escape(scheme).replace(r'\*', '.*')
                     if re.match(regex, url):
                         return endpoint
+        try:
+            timeout = self.registryValue('timeout')
+            response = utils.web.getUrl(url, timeout=timeout)
+            text = response.decode('utf8', errors='replace')
+            match = re.search(
+                r'<link[^>]+?type="application/json\+oembed"[^>]+?href="([^"]+)"',
+                text,
+                re.IGNORECASE)
+            if match:
+                endpoint = match.group(1)
+                endpoint = endpoint.split('?')[0]
+                return endpoint
+        except Exception as e:
+                self.log.debug(f"Failed to discover oEmbed endpoint in HTML: {e}")
         return None
 
     def getOEmbedTitle(self, url):
         """
-        Retrieves the oEmbed title using the providers JSON.
+        Retrieves the oEmbed title.
         """
         try:
             oembed_endpoint = self._getOEmbedEndpoint(url)
             if not oembed_endpoint:
                 return None
- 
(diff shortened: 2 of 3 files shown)


plugins/Web/config.py
```diff
@@ -95,4 +95,12 @@ def configure(advanced):
     seconds the bot will wait for the site to respond, when using the 'fetch'
     command in this plugin. If 0, will use socket.defaulttimeout"""))
 
+conf.registerGlobalValue(Web, 'useOembedRegistry',
+    registry.Boolean(False, _("""Determines whether the bot will use the 
+    oembed.com providers registry.""")))
+
+conf.registerGlobalValue(Web, 'useOembedDiscovery',
+    registry.Boolean(False, _("""Determines whether the bot will use HTML
+    discovery to find oEmbed endpoints.""")))
```
progval, line 104: Name the config variables `plugins.Web.oembed.registry` and `plugins.Web.oembed.discovery`.

Could you also make them channel-specific, but not op-settable?

plugins/Web/plugin.py
```diff
@@ -264,6 +281,55 @@ def url_workaround(url):
...
+                        regex = re.escape(scheme).replace(r'\*', '.*')
+                        if re.match(regex, url):
+                            return endpoint
+        if self.registryValue('useOembedDiscovery'):
+            try:
+                timeout = self.registryValue('timeout')
+                response = utils.web.getUrl(url, timeout=timeout)
+                text = response.decode('utf8', errors='replace')
+                match = re.search(
+                    r'<link[^>]+?type="application/json\+oembed"[^>]+?href="([^"]+)"',
```
progval, line 306: That's insufficient, it does not cover different attribute orders or single quotes.

plugins/Web/plugin.py
```diff
@@ -264,6 +281,55 @@ def url_workaround(url):
...
+                    text,
+                    re.IGNORECASE)
+                if match:
+                    endpoint = match.group(1)
+                    endpoint = endpoint.split('?')[0]
```
progval, line 311: why?

plugins/Web/plugin.py
```diff
@@ -264,6 +281,55 @@ def url_workaround(url):
...
+
+    def getOEmbedTitle(self, url):
+        """
+        Retrieves the oEmbed title.
+        """
+        try:
+            oembed_endpoint = self._getOEmbedEndpoint(url)
+            if not oembed_endpoint:
+                return None
+            oembed_url = f"{oembed_endpoint}?format=json&url={url}"
```
progval, line 325: don't you need to escape the URL? (use `urllib.parse.urlunparse`)

plugins/Web/test.py
```diff
@@ -179,6 +179,31 @@ def testWhitelist(self):
...
+                    'Bacon Lollys')
+            finally:
+                conf.supybot.plugins.Web.useOembedDiscovery.setValue(False)
+
+        def testtitleOembedError(self):
+            try:
+                conf.supybot.plugins.Web.useOembedDiscovery.setValue(True)
+                self.assertError('title https://nonexistent.example.com/post/123')
+            finally:
+                conf.supybot.plugins.Web.useOembedDiscovery.setValue(False)
```
progval, line 205:
```suggestion
        def testtitleOembedRegistry(self):
            with conf.supybot.plugins.Web.useOembedRegistry.context(True):
                self.assertResponse(
                    'title https://www.flickr.com/photos/bees/2362225867/',
                    'Bacon Lollys')

        def testtitleOembedDiscovery(self):
            with conf.supybot.plugins.Web.useOembedDiscovery.context(True):
                self.assertResponse(
                    'title https://flickr.com/photos/bees/2362225867/',
                    'Bacon Lollys')

        def testtitleOembedError(self):
            with conf.supybot.plugins.Web.useOembedDiscovery.context(True):
                self.assertError('title https://nonexistent.example.com/post/123')
```

use the new style, it's shorter

progval (commented):
nice, thanks.

The design looks fine, but i have nitpick belows

PR #1615, opened 2025-03-17
Fix errors reported by msgfmt on Russian translations

mapreri:
Hello,

our QA is warning about these:

```
gettext: locales/ru.po:1922: 'msgid' and 'msgstr' entries do not both begin with '\n'
locales/ru.po:1935: 'msgid' and 'msgstr' entries do not both begin with '\n'
msgfmt: found 2 fatal errors
gettext: plugins/Anonymous/locales/ru.po:101: 'msgid' and 'msgstr' entries do not both begin with '\n'
msgfmt: found 1 fatal error
gettext: plugins/NickAuth/locales/ru.po:29: 'msgid' and 'msgstr' entries do not both begin with '\n'
msgfmt: found 1 fatal error
gettext: plugins/PluginDownloader/locales/ru.po:84: 'msgid' and 'msgstr' entries do not both begin with '\n'
msgfmt: found 1 fatal error
```

Honestly, while looking at simply adding the missing `\n` I noticed that those paragraphs don't seem to respect the original `\n` at all, so I have my doubts those strings are being properly rendered, but personally I was only looking at silencing the warnings :)


progval:
> those paragraphs don't seem to respect the original \n at all, so I have my doubts those strings are being properly rendered

Limnoria strips most newlines before displaying them

PR #1617, opened 2025-04-19
Channel: Make 'invite' command's required capability configurable

progval:
It still requires 'op' by default, but can now be changed with a config value.

This can be useful on servers that do not have a 'free invite' channel mode (like Charybdis/Solanum's +g)

### DIFF
--- a/plugins/Channel/config.py
+++ b/plugins/Channel/config.py
@@ -1,7 +1,7 @@
 ###
 # Copyright (c) 2004-2005, Jeremiah Fincher
 # Copyright (c) 2009, James McCoy
-# Copyright (c) 2010-2021, Valentin Lorentz
+# Copyright (c) 2010-2025, Valentin Lorentz
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


plugins/Channel/config.py
```diff
@@ -63,4 +63,9 @@ def configure(advanced):
         be used (they are optional in the IRC protocol). The standard
         substitutions ($version, $nick, etc.) are all handled appropriately.""")))
 
+conf.registerGroup(Channel, 'invite')
+conf.registerChannelValue(Channel.invite, 'requireCapability',
+    registry.String('op', _("""Determines what capability (if any) the bot should
+    require people trying to use the 'invite' command to have.""")))
```
jlu5, line 69: Should we say explicitly that setting an empty value will allow anyone to use INVITE?

plugins/Channel/test.py
```diff
@@ -414,5 +414,39 @@ def getAfterJoinMessages():
...
+    def testInviteNoCapability(self):
+        self.irc.feedMsg(ircmsgs.op(self.channel, self.nick))
+        m = self.assertError('invite foo',
+                             frm='test!user@with__no_testcap__')
+
+    def testInviteCustomCapability(self):
+        self.irc.feedMsg(ircmsgs.op(self.channel, self.nick))
+
+        self.assertError('invite foo',
+                         frm='test!user@with__no_testcap__')
```
jlu5, line 432: This check duplicates the one in testInviteNoCapability, do we need it?

progval:
I just noticed that it allows people not in a channel to invite themselves, which is counter-intuitive. I'll add a config option to require that the caller be in the channel

progval:
but then ops can't use it to invite themselves back in a channel, hmm...

jlu5 (approved):

progval:
Alternative: a config option (default `op`) for what capability allows one to invite if they are not in channel, and an other (default empty) for what capability allows one to invite if they are

PR #1618, opened 2025-04-20
Remove support for Python 3.6

progval:
Github disabled their Ubuntu 20.04 runners, which means we can't test Python 3.6 in the CI anymore.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,13 +62,6 @@ jobs:
             with-opt-deps: false
             runs-on: ubuntu-22.04
 
-          - python-version: "3.6"
-            with-opt-deps: false
-            runs-on: ubuntu-20.04
-          - python-version: "pypy-3.6"
-            with-opt-deps: false
-            runs-on: ubuntu-20.04
-
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
--- a/setup.py
+++ b/setup.py
@@ -107,8 +107,8 @@ if version:
     fd.write('    pass\n')
     fd.close()
 
-if sys.version_info < (3, 6, 0):
-    sys.stderr.write("Limnoria requires Python 3.6 or newer.")
+if sys.version_info < (3, 7, 0):
+    sys.stderr.write("Limnoria requires Python 3.7 or newer.")
     sys.stderr.write(os.linesep)
     sys.exit(-1)
 
@@ -201,7 +201,6 @@ setup(
         'Operating System :: OS Independent',
         'Operating System :: POSIX',
         'Operating System :: Microsoft :: Windows',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',


jlu5:
FWIW, everything before Python 3.9 is EOL already: https://endoflife.date/python
Do we still need to support those old versions?

For reference:
- Debian 11 (oldstable) has 3.9
- Ubuntu 22.04 (old LTS) has 3.10
- Debian 12 (stable) has 3.11
- Ubuntu 24.04 (current LTS) has 3.12

progval:
I don't see a reason to remove support for these old versions, they don't add a significant burden to us. Maybe just 3.7 so that we can merge #1544?

jlu5:
Removing stale versions would reduce the amount of code paths we have to think about, while having to add compatibility for branches that few users are expected to use creates busywork - I find this demotivating. In the case of #1544 I did not feel like chasing down old Python versions just to verify that my code changes still worked there.

I would like to come up with a longer term plan for declaring which Python versions we support, so that we can be proactive about this, instead of fixing CI only when they drop old versions.

jlu5:
AFAICT we have a handful of version_info checks for 3.7 - 3.9, so it's not like we're never checking the Python version:

```
src/log.py:if sys.version_info >= (3, 8):
src/log.py:if sys.version_info >= (3, 8):
src/utils/gen.py:_OLD_AST = sys.version_info[0:2] < (3, 8)
src/utils/time.py:if sys.version_info >= (3, 9):
src/plugins/Time/test.py:if sys.version_info >= (3, 9):
src/plugins/Time/test.py:    @skipIf(sys.version_info < (3, 7, 0),
```

progval:
So that's only 3 different checks (one for the `validate` arg of `logging.Formatter` in `log.py`, one for `ast` in `gen.py`, and one for the existence of `zoneinfo`) that we'd need to keep after removing 3.6.

> while having to add compatibility for branches that few users are expected to use creates busywork - I find this demotivating

Feel free to ignore that busywork, I'm fine with taking care of it myself including breakages.

One reason I care about not removing support for old Python versions is that, since we don't offer backports for security fixes, I'm afraid people will refrain from updating to a newer Limnoria version if it doesn't support their old Python. (And yes I'm aware that most people who don't update Python also don't update Limnoria)

> In the case of https://github.com/progval/Limnoria/pull/1544 I did not feel like chasing down old Python versions just to verify that my code changes still worked there. (So instead I left it idle for 2 years, which is maybe not great either)

That's why I'm proposing removing support for 3.7 too. And then it would be only the `zoneinfo` check left.

jlu5:
> One reason I care about not removing support for old Python versions is that, since we don't offer backports for security fixes, I'm afraid people will refrain from updating to a newer Limnoria version if it doesn't support their old Python. (And yes I'm aware that most people who don't update Python also don't update Limnoria)

But most people are running Python installed from their distro repository, right? If they're stuck on an unsupported Python version, it's likely they're running an obsolete OS version too. This is not what a responsible sysadmin would do, yet supporting outdated Python versions has the side effect of encouraging people to keep ancient OSes too.

 (Except in the case of Windows, where I guess they can just install a new Python version every few years)

progval:
True, but the risk of running an outdated Limnoria can be higher than running an outdated OS. eg. if Limnoria is running on a NAT raspberry pi, then Limnoria is the only software reachable from the public internet.

jlu5:
> True, but the risk of running an outdated Limnoria can be higher than running an outdated OS. eg. if Limnoria is running on a NATed raspberry pi, then Limnoria is the only software reachable from the public internet.

That might be the case if Limnoria was 100% self contained, but it isn't. If someone is running the bot on an unpatched Python interpreter or OpenSSL, that is hardly a low-risk deployment. It doesn't matter if the bot code is up to date when the rest of your system is fundamentally broken. Presenting things otherwise is offering an illusion of security.

progval:
Python interpreter isn't much of an attack surface. OpenSSL is a good point, though.

progval:
How about this: aim to support all of:

* Debian oldoldstable (if it exists), which [by definition](https://wiki.debian.org/DebianOldOldStable) is still supported under Debian's LTS
* Debian oldstable
* maybe the oldest supported Fedora and Ubuntu? I don't know how these work

However, Debian oldstable (10/buster) has Python 3.7 (as well as an OpenSSL that is still getting security updates) so this would go against your point on codepath complexity

jlu5:
> How about this: aim to support all of:
> 
> * Debian oldoldstable (if it exists), which [by definition](https://wiki.debian.org/DebianOldOldStable) is still supported under Debian's LTS

No, Buster's LTS already expired on June 2024: https://wiki.debian.org/LTS.

> 
> * Debian oldstable
> 

That would be Bullseye and Python 3.9.

> * maybe the oldest supported Fedora and Ubuntu? I don't know how these work
> 

For Ubuntu that would be Ubuntu 22.04 and Python 3.10. Fedora has [fast release cycles](https://endoflife.date/fedora) so  not likely to be a problem here.

There is still [CentOS Stream](https://endoflife.date/centos-stream) and their downstreams like [RHEL](https://endoflife.date/rhel), which have fairly long support cycles:

- RHEL 9 has Python 3.9, but 3.11 and 3.12 are also coinstallable: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/installing_and_using_dynamic_programming_languages/assembly_installing-and-using-python_installing-and-using-dynamic-programming-languages
- RHEL 8 has Python 3.6 :(  CentOS Stream 8 is already EOL, but RHEL 8 isn't. There are alternatives like Alma Linux and Rocky Linux which indicate the same support cycle as RHEL 8, but I don't know what their patching strategy is now that [RHEL has stopped offering source code](https://almalinux.org/blog/impact-of-rhel-changes/).

Anyways, if we're intent on dropping 3.6 I think Python 3.9 makes sense as the next baseline. This roughly lines up with:
- Debian oldstable
- Oldest supported Ubuntu LTS
- Oldest supported CentOS Stream release

This should reasonably cover most people's use cases.

progval:
> No, Buster's LTS already expired on June 2024: https://wiki.debian.org/LTS.

Ah right, I misread the LTS page

> RHEL 8 has Python 3.6 :( CentOS Stream 8 is already EOL, but RHEL 8 isn't. There are alternatives like Alma Linux and Rocky Linux which indicate the same support cycle as RHEL 8

Almalinux 9.x is the only supported branch, and they have Python 3.9. Almalinux 8.x had Python 3.6 but its support ended a year ago.

Rockylinux 8.10 has Python 3.6 and is said to be supported until 2029, and Rockylinux 9.x has Python 3.9.

So a semi-formal policy, we can decide to support all of:

* Debian oldoldstable if it exists (currently it does not)
* Debian oldstable, stable, testing, unstable
* all currently supported Ubuntu, Fedora, and CentOS Stream versions
* Arch
* latest Almalinux/Rockylinux branch
* any supported Almalinux/Rockylinux branch if it's not too inconvenient for us

> Anyways, if we're intent on dropping 3.6 I think Python 3.9 makes sense as the next baseline.

I agree

jlu5:
> So a semi-formal policy, we can decide to support all of:
> 
>    * Debian oldoldstable if it exists (currently it does not)
> 
>    * Debian oldstable, stable, testing, unstable
> 
>    * all currently supported Ubuntu, Fedora, and CentOS Stream versions
> 
>    * Arch
> 
>    * latest Almalinux/Rockylinux branch

This sounds fine to me. I purposely simplified the discussion to only define the *oldest* supported version, as obviously anything between that and the latest Python branch should work as well.

So the target could then be rephrased as:

- All supported (upstream) Python releases, plus:
  - The default Python version available in all active Debian and Ubuntu LTS releases
  - The default Python version available in the latest RHEL release

> 
>    * any supported Almalinux/Rockylinux branch if it's not too inconvenient for us
> 

I don't think we need to mention anything that's only supported on a best effort basis. Inconvenient is subjective and doesn't give anyone reading such a policy concrete information. The (current) minimum Python version should be listed in the docs anyways.

progval:
> Inconvenient is subjective and doesn't give anyone reading such a policy concrete information. The (current) minimum Python version should be listed in the docs anyways.

I meant these as a guideline for us to decide what versions we support. And then of course, write these versions in the doc

progval:
Replaced by #1625

PR #1623, opened 2025-04-27
Google: remove everything other than translate

jlu5:
This has been deprecated since 2016 and is largely broken now.


progval:
We should provide an alternative repository for a smaller plugin with the `@translate` command before removing the plugin

plugins/Google/test.py
```diff
@@ -34,34 +34,7 @@
...
-            self.assertRegexp('google foo', 'https?://.*')
-            self.assertNotRegexp('google foo', '<https?://.*>')
-
-        def testSearchOneToOne(self):
-            self.assertRegexp('google dupa', ';')
-            self.assertNotError('config plugins.Google.oneToOne True')
-            self.assertNotRegexp('google dupa', ';')
-
         def testTranslate(self):
-            self.assertRegexp('translate en es hello world', 'Hola mundo')
```
progval, line 65: Why remove the translation check?

jlu5: Google now returns "Hola Mundo" instead of "Hola mundo" which means the test was actually failing. The output is subject to change, so I'm not sure we should hardcode it.

progval: `(?i)` should make it case-insensitive

jlu5: Actually no, when I run the test locally, it shows "Hola World". While on my production bot (hosted elsewhere) it shows "Hola Mundo"?

progval: :shrug: let's allow both options

jlu5: I don't get it. The website shows multiple possible outputs too, but "Hola World" seems plainly wrong. Yet 
I can't reproduce that output with wget ...

image

If we can't be sure what the API will return I don't think we should hardcode any expectation at all.

progval: but at least it checks we're not returning complete nonsense. And it doesn't run on the CI anyway

jlu5: Ok I'll just leave it in even though it doesn't pass for me

PR #1625, opened 2025-05-03
Bump minimum Python version to 3.9

jlu5:


progval:
See https://github.com/progval/Limnoria/pull/1618 for the discussion about what Python versions we want to support

PR #1629, opened 2025-05-27
NickAuth: Make username argument to 'nick add' required

progval:
optional() is greedy, so if someone uses 'nickauth nick add MyNick' and MyNick is also a bot account name, the command fails. This is very confusing.


jlu5:
Is there no nice way to preserve the current behaviour? I think the worst case implementation would be to manually count whether we've been passed one or two arguments. Even if it's not so idiomatic, it might be better than making the commands less flexible

progval:
good idea

PR #1642, opened 2025-09-12
CI: Update 3.14 from alpha1 to rc2

progval:
cffi 2.0 was published this week and broke support for some 3.14 alpha versions. This caused a segfault when trying to use python3-cryptography for ecdsa


jlu5:
There's another test failure because URLopener has been removed in Python 3.14. But we can fix this separately so LGTM

jlu5 (approved):

PR #1657, opened 2026-03-08
Add Windows CI + fix a bunch of bugs/tests on Windows

progval:


plugins/GPG/plugin.py
```diff
@@ -221,6 +221,12 @@ def auth(self, irc, msg, args, url):
                 content = content.decode()
             match = self._auth_re.search(content)
             if not match:
+                import base64
+                print('GPG_DEBUG url=%r' % url, flush=True)
+                print('GPG_DEBUG content_repr=%r' % content, flush=True)
+                print('GPG_DEBUG content_b64=%s' % base64.b64encode(
+                        content.encode('utf-8', errors='replace')
+                        ).decode(), flush=True)
```
jlu5: Should these be log.debug instead?

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,44 @@
...
+          python3 -m pip install --upgrade pip setuptools
+
+      - name: Install optional dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          pip3 install -r requirements.txt
+
+      - name: Install
+        run: |
+          set LIMNORIA_WARN_OLD_PYTHON=0
```
jlu5: I think this env var can be removed

src/utils/file.py
```diff
@@ -39,8 +39,8 @@
 from . import crypt, force
 
 def sanitizeName(filename):
-    """Removes / from filenames and escapes them if they are '.' or '..'."""
-    filename = filename.replace('/', '')
+    r"""Removes '/', '\', and '|' from filenames and escapes them if they are '.' or '..'."""
```
jlu5, line 42: For consistency, should we do the same for [all reserved chars](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions) on Windows `<>:"/\|?*`?

This feels quite restrictive though, so maybe there should be a config toggle which only takes effect outside of Windows.

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,44 @@
...
+          pip install .
+
+      - name: Test with unittest
+        run: |
+          supybot-test test -v --plugins-dir=./plugins/ --no-network --exclude="*Fediverse" --exclude="*GPG" --exclude="*RSS"
+
+      - name: Test with irctest
+        if: "${{ matrix.with-opt-deps && matrix.python-version != '3.7' && matrix.python-version != 'pypy-3.7' && matrix.python-version != 'pypy-3.9' }}"
```
jlu5: I don't think we support 3.7 anymore or test PyPy on Windows. In fact, I don't see a matrix of Python versions defined here at all, so I assume it's just using GitHub's default?

src/utils/file.py
progval: We need the same rules for people who want to switch between Linux and Windows

jlu5 (approved):

PR #1669, opened 2026-04-11
registry: Add 'settable' argument to make some config variables read-only

progval:
This removes the special-case in the Config plugin for variables that should not be changed when supybot.commands.allowShell is False

I'm not fond of the name. Should I rename it to something like `readOnly`?

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -85,15 +85,17 @@ def registerGroup(Group, name, group=None, **kwargs):
         group = registry.Group(**kwargs)
     return Group.register(name, group)
 
-def registerGlobalValue(group, name, value):
+def registerGlobalValue(group, name, value, settable=True):
     value._networkValue = False
     value._channelValue = False
+    value._settable = settable
     return group.register(name, value)
 
-def registerNetworkValue(group, name, value):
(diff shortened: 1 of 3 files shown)


src/conf.py
```diff
@@ -868,7 +871,8 @@ class ValidBrackets(registry.OnlySomeStrings):
     to prevent MITM from the IRC network itself (vulnerable IRCd or IRCops)
     from gaining shell access to the bot's server by impersonating the owner.
     Setting this to False also disables plugins and commands that can be
-    used to indirectly gain shell access.""")))
+    used to indirectly gain shell access.""")),
+    settable=lambda: supybot.commands.allowShell())
```
jlu5, line 875: Do we need to wrap this in a lambda?

LGTM otherwise

progval: we don't, but I think it's clearer this way that we're actually passing a callable this way. I don't have strong feelings about it, though.

jlu5: Hmm, isn't this what type hints are for in modern codebases :D

progval: yes :(

PR #1672, opened 2026-04-25
Revert "Services: Fix some networks' nicks not being registered in the config"

luk1337:
do001() is called too late and it ends up deleting NickServ.password.%s from config file.

This reverts commit 026dbf314aac459466264ceb97b1ca7c83c497fb.

PR #1673, opened 2026-04-25
Services: Fix config initialization

progval:
Fixes 026dbf314aac459466264ceb97b1ca7c83c497fb which never actually ran as there is already a method named do001 defined later in the file, which overrode this.

Additionally, do001 is not the right place to initialize, because __call__ is called for earlier messages (connection registration) and needs to access the config values

Closes #1672


luk1337 (approved):
works for me

progval:
thanks!

